PR #437, opened 2021-12-06
feat: add lazy install option to package config

uesyn:
If we want to install some applications normally and the others applications lazily, we must manage two configurations and run a shell script like the following:

````
$ aqua --config=/path/to/aqua.yaml install
$ aqua --config=/path/to/aqua-lazy.yaml install -l
````

And we need to export AQUA_GLOBAL_CONFIG to refer two config files.

```
export AQUA_GLOBAL_CONFIG=/path/to/aqua.yaml:/path/to/aqua-lazy.yaml
````

This configuration and script is a bit ugly. In order to be able to manage the installation all together in one configuration, I would like to add the following lazy install option.

```
packages:
- name: golang/go
  version: "1.16
  lazy: true
``` 

If `lazy: true` is set in a package as above, `aqua install` only creates links and doesn't download, even if no `-l` option.

### DIFF
--- a/pkg/controller/config.go
+++ b/pkg/controller/config.go
@@ -16,10 +16,11 @@ import (
 )
 
 type Package struct {
-	Name     string `validate:"required"`
-	Registry string `validate:"required" yaml:",omitempty"`
-	Version  string `validate:"required" yaml:",omitempty"`
-	Import   string `yaml:",omitempty"`
+	Name        string `validate:"required"`
+	Registry    string `validate:"required" yaml:",omitempty"`
+	Version     string `validate:"required" yaml:",omitempty"`
+	Import      string `yaml:",omitempty"`
+	LazyInstall bool   `yaml:"lazy,omitempty"`
 }
 
 func (pkg *Package) UnmarshalYAML(unmarshal func(interface{}) error) error {
--- a/pkg/controller/install_packages.go
+++ b/pkg/controller/install_packages.go
@@ -63,13 +63,20 @@ func (ctrl *Controller) installPackages(ctx context.Context, cfg *Config, regist
 
 	for _, pkg := range cfg.Packages {
 		go func(pkg *Package) {
-			defer wg.Done()
 			maxInstallChan <- struct{}{}
+			defer func() {
+				wg.Done()
+				<-maxInstallChan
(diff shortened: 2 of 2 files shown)


suzuki-shunsuke:

The discussion is needed.

We would like to keep aqua.yaml simple, and we have to be careful when we add features in aqua.yaml.

> we want to install some applications normally and the others applications lazily

Could you explain why some packages should be install immediately?
Maybe there is the other options to solve your problem.

> If lazy: true is set in a package as above, aqua install only creates links and doesn't download, even if no -l option.

If we add `lazy: true`, we may want `--force` option to ignore `lazy: true`, but we don't want to add options as much as possible.

uesyn:
> The discussion is needed.
> We would like to keep aqua.yaml simple, and we have to be careful when we add features in aqua.yaml.

I should have created the Issue first.... I'm sorry about that 🙏 

> Could you explain why some packages should be install immediately?
> Maybe there is the other options to solve your problem.

I often develop on an ephemeral development environments.
Sometimes it's on a VM, sometimes it's on a github codespace.
In the case of codespace in particular, I have no control over the initial state of the devcontainer that is used in repositories other than my own.
To solve this problem, I use aqua to install the applications I need, but it takes a long time to bootstrap if install everything.
So I install the applications I use regularly on bootstrap, and install the others lazily.

> If we add `lazy: true`, we may want --force option to ignore `lazy: true`, but we don't want to add options as much as possible.

hmm...
In the current implementation, the installation unit is a file, which makes install inflexible. If it were possible to split the installation target within a single config file, my problem would be solved, and maybe this related to the following Issue.
https://github.com/aquaproj/aqua/issues/228

How about adding an attribute like "label" in package config to specify the target application?
If not specified, all will be targeted, and if labels are specified, only applications which match the labels will be targeted.

uesyn:
I understood your aqua development policy.
I also think that simplicity is important 👍 

> So would you create an issue about this?

created!
https://github.com/aquaproj/aqua/issues/441

I will close this PR.
Thank you so much for taking the time and your kind answer 🙏

suzuki-shunsuke:
Thank you for your understanding and creating the issue!

PR #519, opened 2022-01-08
fix: treat aqua.yaml which places on init as normal file rather than executable

pankona:
Change file mode of aqua.yaml (which is placed on `aqua init`) from 755 to 644 since it is just a configuration file but not executable 💪


suzuki-shunsuke:

pankona:
Thank you! 😋

PR #859, opened 2022-06-13
feat: add 'completion bash|zsh' commands

### LINKED ISSUE #856: Shell completion for aqua commands
## Feature Overview

Please provide a command like `aqua completion bash` to generate the shell completion script for each shell. This seems to be easily achievable with:
https://cli.urfave.org/v2/#bash-completion

## Why is the feature needed?

> Please explain the problem you want to solve.

Many cli tools provide `completion <shell>` command for users. It's highly convenient to have it with aqua too.

## Does the feature include Breaking Changes?

I don't think so.

## Example Code

> command and configuration

Once it's provided, users would run a command like this to set up their shell completions:

```console
$ source <(aqua completion bash)
```

Thank you!

tadayosi:
Fix #856

With this fix, now you can add the following line to `.bashrc` to enable commands completion for aqua.
```
if which aqua &> /dev/null; then source <(aqua completion bash); fi
```

### DIFF
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -57,6 +57,11 @@ jobs:
     - run: github-compare -v
     - run: terrafmt version
 
+    - name: output bash completion
+      run: aqua completion bash
+    - name: output zsh completion
+      run: aqua completion zsh
+
     - run: aqua g -i suzuki-shunsuke/tfcmt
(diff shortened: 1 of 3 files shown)


suzuki-shunsuke:
Thank you for your contribution.

suzuki-shunsuke (approved):

PR #794, opened 2022-05-25
feat: support verifying downloaded assets with checksums

suzuki-shunsuke:
https://github.com/aquaproj/aqua/issues/427

Issue #427, opened 2021-12-04
Verify the checksum


suzuki-shunsuke:
ref. https://zenn.dev/shunsuke_suzuki/scraps/7cfc2d3a5c6d04

## Overview

Verify the checksum of downloaded file, and if the checksum is wrong make the installation failure.

## Motivation

Make aqua secure.
Prevent the supply chain atack.

## Consideration

We have to keep aqua simple. We should avoid making aqua complicated by introducing checksum verification.

## Proposal of Specification

When a tool is installed, aqua verifies the checksum as the following.

1. Download the file in the temporal directory
1. Calculate the checksum from the downloaded file
1. Read the file `.aqua-checksums.json` on the same directory as aqua configuration file. If `.aqua-checksums.json` isn't found, aqua treats the file is empty
1. Get the expected checksum of downloaded file from `.aqua-checksums.json`
1. If the actual checksum is different from the expected checksum, make the installation failure. If the checksum isn't found in `.aqua-checksums.json`, the actual checksum is added to `.aqua-checksums.json`
1. Install the file

## :warning: Note

If the file is falsified before the correct checksum is added to `.aqua-checksums.json`, it is difficult to detect the falsification.

## :bulb: Ignore the specific checksum verification

If you would like to ignore the specific checksum verification, you can do it by removing the checksum from `.aqua-checksums.json`.

## .aqua-checksums.json

Pairs of package id and checksum are recorded.

e.g.

```json
{
  "github_archive/github.com/tfutils/tfenv/v2.2.3": "0b42330aeed675ad3e5dd6063dbd1daabecb36180e515cca1c6e105dd7a1fa49"
}
```

This file is created and updated by `aqua` automatically, so you don't have to update this file manually.

## Question: Should `.aqua-checksums.json` be managed with Git?

Coming soon

## Algorithm to calculate checksums

sha256 is used for goreleaser by default, so aqua would also support sha256 by default.
In future, aqua may support other algorithms too.

## Idea: command to update `.aqua-checksums.json` like `terraform providers lock`

Coming soon

## Future work

* Verify Registry
* Verify aqua-proxy
* Support checksum algorithm other than sha256
* Get checksums from GitHub Releases's file

## Implementation

* https://github.com/aquaproj/aqua/pull/794

## Reference

* https://news.ycombinator.com/item?id=29437903
* https://pkg.go.dev/crypto/sha256

suzuki-shunsuke:
[...]

PR #1030, opened 2022-07-21
feat: support verifying downloaded assets with checksums

suzuki-shunsuke:
#427 Follow up #794


suzuki-shunsuke:
https://github.com/aquaproj/aqua/pull/1030

Issue #1041, opened 2022-07-27
Support `update-aqua` command


suzuki-shunsuke:
## Feature Overview

Add a subcommand `update-aqua`, which installs or updates aqua.

## Why is the feature needed?

> Please explain the problem you want to solve.

To make the update of aqua easy.
I'd like users to use the latest version, but some users use the old version.
The old version sometimes causes trouble.

## Does the feature include Breaking Changes?

No.

## Example Code

> command and configuration

```console
$ aqua update-aqua [-o <install or update path>] [version]
```

By default, `aqua update-aqua` updates aqua to the latest version.

```console
$ aqua update-aqua
```

You can specify the version and the path of aqua.

```console
$ aqua update-aqua -o ~/bin/aqua v1.17.0
```

suzuki-shunsuke:
https://github.com/aquaproj/aqua/pull/1060

suzuki-shunsuke:
I've changed the specification.

# Specification

```console
$ aqua update-aqua [version]
```

By default, the latest version is installed.
aqua is installed in `$AQUA_ROOT_DIR/pkgs` same as other packages and a symbolic link is created in `$AQUA_ROOT_DIR/bin/aqua`.

⚠️ In Windows, a bat file and a shell script are created in `$AQUA_ROOT_DIR/bat` and `$AQUA_ROOT_DIR/bin` instead of a symbolic link.

You can't specify the install path. I think you don't have to change the install path.

e.g.

```console
$ aqua update-aqua
```

PR #1060, opened 2022-07-30
feat: add update-aqua command

suzuki-shunsuke:
Close https://github.com/aquaproj/aqua/issues/1041


suzuki-shunsuke:
I don't know the reason but it doesn't work well on darwin/arm64. 🤔 

```console
$ ~/go/bin/aqua update-aqua
INFO[0001] updating aqua                                 aqua_version= env=darwin/arm64 new_version=v1.17.1 program=aqua

$ aqua -v
zsh: killed     aqua -v
```

Issue #1289, opened 2022-10-09
create a temporal file: open ***: is a directory


suzuki-shunsuke:
## aqua version

Please use the latest version.

```console
$ aqua -v
aqua version 1.20.0-3 (29e4189957ccce0ec3262fd6942045c244505cc8)
```

## Environment

* OS (Windows, Linux, macOS, etc): darwin
* CPU Architecture (amd64, arm64, etc): arm64

## How to reproduce

* aqua.yaml
* executed command and output

> aqua.yaml should be not partial but complete configuration.
> Please remove unneeded configuration to reproduce the issue.

https://github.com/aquaproj/aqua-registry/blob/0324a2ee020f7f151a717741e1420e499e3e7424/aqua-all.yaml

aqua-all.yaml

```yaml
---
# aqua - Declarative CLI Version Manager
# https://aquaproj.github.io/
checksum:
  enabled: true
registries:
  - name: standard
    type: local
    path: registry.yaml
packages:
  - import: pkgs/*/*/pkg.yaml
  - import: pkgs/*/*/*/pkg.yaml
  - import: pkgs/*/*/*/*/pkg.yaml
  - import: pkgs/*/*/*/*/*/pkg.yaml
```

```console
$ aqua i
```

* b3nj5m1n/xdg-ninja@v0.2.0.1 github_archive
* bats-core/bats-core@v1.8.0 github_archive
* npryce/adr-tools@3.0.0 github_archive
* shellspec/shellspec@0.28.1 github_archive

```
ERRO[0019] install the package                           aqua_version=1.20.0-3 env=darwin/arm64 error="create a temporal file: open /var/folders/lk/bw_hsbdd72l9ckd0tz61xz2m0000gn/T/349031091: is a directory" package_name=b3nj5m1n/xdg-ninja package_version=v0.2.0.1 program=aqua registry=standard temp_file=/var/folders/lk/bw_hsbdd72l9ckd0tz61xz2m0000gn/T/349031091

ERRO[0020] install the package                           aqua_version=1.20.0-3 env=darwin/arm64 error="create a temporal file: open /var/folders/lk/bw_hsbdd72l9ckd0tz61xz2m0000gn/T/562990198: is a directory" package_name=bats-core/bats-core package_version=v1.8.0 program=aqua registry=standard temp_file=/var/folders/lk/bw_hsbdd72l9ckd0tz61xz2m0000gn/T/562990198

[...]

PR #1290, opened 2022-10-09
fix: fix assetName

suzuki-shunsuke:
Close https://github.com/aquaproj/aqua/issues/1289

If package type is `github_archive`, `param.AssetName` is empty.
`filepath.Base("")` returns `"."`.


pkg/installpackage/checksum.go
```diff
@@ -95,7 +95,7 @@ func (inst *Installer) verifyChecksum(ctx context.Context, logE *logrus.Entry, p
 	// Compare the checksum
 	// Store the checksum to aqua-checksums.json
 
-	assetName := filepath.Base(param.AssetName)
```
suzuki-shunsuke, line 98: I'm not sure why `filepath.Base` is used. 🤔

PR #1413, opened 2022-11-16
breaking changes for v2

suzuki-shunsuke:
- https://github.com/aquaproj/aqua/issues/1408
- https://github.com/aquaproj/aqua/issues/1409
- https://github.com/aquaproj/aqua/issues/1411
- ~https://github.com/aquaproj/aqua/issues/1412~
- https://github.com/aquaproj/aqua/issues/1410
- https://github.com/aquaproj/aqua/issues/1404

Issue #1399, opened 2022-11-12
Road to v2


suzuki-shunsuke:
Major update means breaking changes.
We should avoid breaking changes as much as possible, but we would like to change some points.

https://github.com/aquaproj/aqua/milestone/96

suzuki-shunsuke:
- https://github.com/aquaproj/aquaproj.github.io/pull/248

Issue #1411, opened 2022-11-15
Remove `supported_if`


suzuki-shunsuke:
## What is changed

Remove [supported_if](https://aquaproj.github.io/docs/reference/registry-config/supported-if).

## Why this change is needed

From aqua v1.12.0 or later, we strongly recommend [supported_envs](https://aquaproj.github.io/docs/reference/registry-config/supported-envs) instead of supported_if. supported_envs is simpler than supported_if, and better in terms of the performance.

Issue #1410, opened 2022-11-15
When AQUA_CONFIG or -c option is set, aqua shouldn't search configuration files


suzuki-shunsuke:
https://aquaproj.github.io/docs/reference/config#configuration-file-path

## What is changed

When AQUA_CONFIG or -c option is set, aqua shouldn't search configuration files.

## Why this change is needed

When we specify the configuration file path explicitly, we would like to use the configuration file and shouldn't use other configuration files.
The current behaviour is a bit confusing and causes unneeded package installation.

PR #1516, opened 2023-01-03
breaking changes for v2

suzuki-shunsuke:
- https://github.com/aquaproj/aqua/issues/1399
- https://github.com/aquaproj/aqua/pull/1413


suzuki-shunsuke:
https://github.com/aquaproj/aqua/releases/tag/v2.0.0-1

suzuki-shunsuke:
v2.0.0-2

suzuki-shunsuke:
# Release Note (Draft)

This release includes several breaking changes.
You may have to fix code to upgrade aqua to v2, but it would be easy.
Most important change is that **only standard registry is allowed by default**.
This breaking change makes aqua secure by default.
Other changes are not so important as the registry restriction, but these changes makes aqua simple.

## ⚠️ Breaking Changes

#1404 #1413 Only standard registry is allowed by default
#1410 #1413 When `AQUA_CONFIG` or `-c` option is set, aqua shouldn't search configuration files
#1409 #1413 Fix the default `files[].name`

Removed some features.

#1408 #1413 Remove the package type [go](https://aquaproj.github.io/docs/reference/registry-config/go-package)
#1411 #1413 Remove [supported_if](https://aquaproj.github.io/docs/reference/registry-config/supported-if)

## Only standard registry is allowed by default

#1404 #1413

If you don't set policies, aqua uses the default policy allowing only standard registries.
If you set policies the default policy is ignored.

### Why this change is needed

This change makes aqua secure by default.
Almost users don't need registries other than standard registry.
Other registries are useful but also can be abused.

### How to migrate

If you use only standard registry, you don't have to do anything.
Otherwise you have to set policies to allow registries other than standard registry.
About policies, please see the document.

https://aquaproj.github.io/docs/tutorial-extras/policy-as-code/

## When `AQUA_CONFIG` or `-c` option is set, aqua shouldn't search configuration files

#1410 #1413

### Why this change is needed

When we specify the configuration file path explicitly, we would like to use the configuration file and shouldn't use other configuration files.
The current behaviour is a bit confusing and causes unneeded package installation.

## Fix the default `files[].name`

#1409 #1413 

If the package has a `name` field, the `name` is split with `/` and the last element is used as the default file name.

For example, please see the following example.

```yaml
name: cert-manager/cert-manager/cmctl
repo_owner: cert-manager
repo_name: cert-manager
```

Then in aqua v1 the default setting of `files` is the following.

```yaml
files:
- name: cert-manager
```

On the other hand, in aqua v2 the default setting of `files` is the following.

```yaml
files:
- name: cmctl
```

### Why this change is needed

We think aqua v2's default setting is better than aqua v1 in many cases.

### How to migrate

If you maintain registries, you might have to fix them.
And you may have to update standard registry to v4.0.0 or later.

## Remove the package type `go`

#1408 #1413

### Why this change is needed

To make aqua simple.
The package type `go` is left only for the compatibility.
You should use [go_install](https://aquaproj.github.io/docs/reference/registry-config/go-install-package) instead of `go` package.

## Remove supported_if

#1411 #1413

### Why this change is needed

From aqua v1.12.0 or later, we strongly recommend [supported_envs](https://aquaproj.github.io/docs/reference/registry-config/supported-envs) instead of supported_if. supported_envs is simpler than supported_if, and better in terms of the performance.

Issue #1546, opened 2023-01-09
Migrate hashicorp/go-version to another library such as `Masterminds/semver`


suzuki-shunsuke:
## Feature Overview

Migrate [hashicorp/go-version](https://github.com/hashicorp/go-version) to another library such as [Masterminds/semver](https://github.com/Masterminds/semver) and [coreos/go-semver](https://github.com/coreos/go-semver).

## Why is the feature needed?

> Please explain the problem you want to solve.

hashicorp/go-version has a bug regarding to the comparison of prerelease versions.
This is obviously a bug and there was [a bug report](https://github.com/hashicorp/go-version/issues/101), but maintainers closed the issue and seems not to fix it.
So I'd like to migrate hashicorp/go-version to another libraries.

But I have a concern. There are several alternatives, but hashicorp/go-version is most popular so I'm not sure we could (should) migrate hashicorp/go-version to another libraries.

## Does the feature include Breaking Changes?

Probably no.
This is an internal refactoring and we will keep the compatibility as much as possible, but maybe the behaviour of `version_constraint` would be changed.

## Example Code

> command and configuration

```console
$ 
```

```yaml

```

## Reference

- https://zenn.dev/shunsuke_suzuki/scraps/e418b0039b232a
- https://semver.org/#spec-item-11

suzuki-shunsuke:
Oh, my god.

Masterminds/semver has a same bug.

```go
package main

import (
	"fmt"
	"log"

	"github.com/Masterminds/semver/v3"
)

const (
	ver        = "1.1.0-alpha"
	constraint = ">= 1.0.0"
)

func main() {
	if err := core(); err != nil {
		log.Fatal(err)
	}
}

func core() error {
	a, err := semver.NewConstraint(constraint)
	if err != nil {
		return err
	}
	ver, err := semver.NewVersion(ver)
	if err != nil {
		return err
	}
	f := a.Check(ver)
	fmt.Println(f)
	return nil
}
```

```console
$ go run .
false
```

suzuki-shunsuke:
coreos/go-semver isn't tagged for a long time.

https://github.com/coreos/go-semver/issues/41

suzuki-shunsuke:
- https://github.com/aquaproj/aqua/issues/1573

PR #1551, opened 2023-01-10
fix: replace hashicorp/go-version to Masterminds/semver

suzuki-shunsuke:
- https://github.com/aquaproj/aqua/issues/1546


suzuki-shunsuke:
- https://github.com/aquaproj/aqua/issues/1546#issuecomment-1377010750

PR #1559, opened 2023-01-13
fix: lock installing cosign

suzuki-shunsuke:
## bug: Lock installing Cosign

When aqua tries to install cosign multiple times in parallel, it would cause the error.
Cosign should be installed only once, and another goroutines should wait until the installation finishes.

## bug: Fix a bug that options of Cosign could be wrong if the same package's multiple versions are installed at the same time


suzuki-shunsuke:
https://github.com/aquaproj/aqua/releases/tag/v1.31.0-3

suzuki-shunsuke:
v1.31.0-5

PR #1798, opened 2023-03-23
feat: support dmg format 

### LINKED ISSUE #1706: Is it possible that aqua will support .dmg files in the future?
## Feature Overview
I want to make tools in .dmg files available in aqua

## Why is the feature needed?

> Please explain the problem you want to solve.
>
Some cli tools are distributed in darwin as .dmg files only. (For example, [in the case of AWS-Vault](https://github.com/99designs/aws-vault/releases/tag/v7.0.1)).
If an aqua user requests the cli author to distribute binaries for darwin, the author will say, "Why should I release a duplicate package when I distribute a dmg file for darwin?" This would be the case. I too do not think this solution is reasonable.
Maybe the best way to make everyone happy in such a case is to use a version manager like aqua to handle it.

I think it is technically possible to support dmg files, since dmg files themselves can be handled by brew cli and so on.
If I make a pull request, is it possible for aqua to accept this change?

## Does the feature include Breaking Changes?
No, it is only an additional feature to support .dmg extensions, so aqua does not expect it to affect others

## Example Code

```

### PR DESCRIPTION
**Pull request by u-yas**
close #1706 

Enabled the use of dmg format (mac os only) files in aqua install.

The dmg format file is solved and made available as follows.

1. run hdiutil attach command and mount the dmg file
2. store the files and folders under the mounted folder in the specified aqua pkgs folder
3. execute hdiutil detach command to detach the mounted folder

If you have mac os, try editing the following files under the aqua folder (aws-vault only provides dmg files for mac)

```yml
# aqua/registry.yaml
packages:
  - type: github_release
    repo_owner: 99designs
    repo_name: aws-vault
    description: A vault for securely storing and accessing AWS credentials in development environments
    asset: aws-vault-{{.OS}}-{{.Arch}}
    format: raw
    supported_envs:
      - linux
      - windows/arm64
      - darwin
    checksum:
      type: github_release
      asset: SHA256SUMS
      file_format: regexp
      algorithm: sha256
      pattern:
        checksum: "^(\\b[A-Fa-f0-9]{64}\\b)"
        file: "^\\b[A-Fa-f0-9]{64}\\b\\s+(\\S+)$"
    overrides:
      - goos: darwin
        format: dmg
        asset: aws-vault-darwin-{{.Arch}}.{{.Format}}
    version_constraint: semver(">= 6.5.0")
    version_overrides:
      - version_constraint: semver(">= 5.2.0")
        supported_envs:
          - linux
      - version_constraint: semver(">= 5.1.0")
        supported_envs:
          - linux
        checksum:
          enabled: false
      - version_constraint: semver(">= 4.7.0")
        supported_envs:
          - linux/amd64
        checksum:
          enabled: false
      - version_constraint: semver(">= 3.5.0")
        supported_envs:
          - linux/amd64
          - darwin
        rosetta2: true
        checksum:
          enabled: false
      - version_constraint: semver(">= 3.0.0-dev")
        replacements:
          amd64: x86_64
          darwin: Darwin
        supported_envs:
          - darwin
        rosetta2: true
        checksum:
          enabled: false
      - version_constraint: semver(">= 2.4.0")
        replacements:
          amd64: x86_64
          darwin: Darwin
          linux: Linux
        supported_envs:
          - linux/amd64
          - darwin
        rosetta2: true
        checksum:
          enabled: false
      - version_constraint: semver(">= 2.3.3")
        replacements:
          amd64: x86_64
          darwin: Darwin
        supported_envs:
          - darwin
        rosetta2: true
        checksum:
          enabled: false
      - version_constraint: semver(">= 2.3.2")
        replacements:
          amd64: x86_64
          darwin: Darwin
          linux: Linux
        supported_envs:
          - linux/amd64
          - darwin
        rosetta2: true
        checksum:
          enabled: false
      - version_constraint: semver("< 2.3.2")
        replacements:
          amd64: x86_64
          darwin: Darwin
        supported_envs:
          - darwin
        rosetta2: true
        checksum:
          enabled: false
```

```yml
# aqua/imports/aws-vault.yml
packages:
  - name: 99designs/aws-vault@v7.1.2
    registry: local

```

```yml
# aqua/policy.yaml
---
# aqua Policy
# https://aquaproj.github.io/docs/tutorial-extras/policy-as-code
registries:
- type: standard
  ref: semver(">= 3.0.0")
- name: main
  type: local
  path: ../tests/main/registry.yaml
- name: cosign
  type: local
  path: ../tests/cosign/registry.yaml
- name: slsa
  type: local
  path: ../tests/slsa/registry.yaml
- name: local
  type: local
  path: registry.yaml
packages:
- registry: standard
- registry: main
- registry: cosign
- registry: slsa
- registry: local

```

```yml
# aqua/aqua.yaml
---
# aqua - Declarative CLI Version Manager
# https://aquaproj.github.io/
checksum:
  enabled: true
  require_checksum: true
registries:
  - type: standard
    ref: v3.148.0 # renovate: depName=aquaproj/aqua-registry
  - name: local
    type: local
    path: registry.yaml
packages:
  - import: imports/*.yaml

```

And go run . /cmd/aqua/main.go update-checksum command to update the checksum, then go run . /cmd/aqua/main.go i . After that, you will see that aws-vault is running successfully!

https://user-images.githubusercontent.com/52757302/227594444-a53e583b-e7f3-4fff-a1a5-67b474ef99a0.mp4

## Concerns.
This feature is only for mac os, do we need something like darwin-test.yaml as well as windows-test.yaml? If needed, add gha.


suzuki-shunsuke:

suzuki-shunsuke (commented):
The result of `aqua gr` is wrong.

```console
$ go run ./cmd/aqua gr 99designs/aws-vault
packages:
  - type: github_release
    repo_owner: 99designs
    repo_name: aws-vault
    description: A vault for securely storing and accessing AWS credentials in development environments
    asset: aws-vault-{{.OS}}-{{.Arch}}
    format: raw
    supported_envs:
      - linux
      - windows/arm64
    checksum:
      type: github_release
      asset: SHA256SUMS
      file_format: regexp
      algorithm: sha256
      pattern:
        checksum: "^(\\b[A-Fa-f0-9]{64}\\b)"
        file: "^\\b[A-Fa-f0-9]{64}\\b\\s+(\\S+)$"
```

pkg/asset/format.go
```diff
@@ -3,15 +3,27 @@ package asset
...
 )
 
-const formatRaw = "raw"
+const formatRaw string = "raw"
+
+// mholt/archiver/v3 not support but aqua support
+func aquaSupportFormat(assetName string) string {
+	formatOtherFormats := map[string]struct{}{unarchive.FormatDMG: {}, formatRaw: {}}
+
+	ext := strings.TrimPrefix(assetName, ".")
```
suzuki-shunsuke, line 16: It looks strange at a glance.

suzuki-shunsuke:
I added a test case for dmg. We have to fix to pass the test case.
https://github.com/aquaproj/aqua/pull/1798/commits/6c19fb4fe8fca354f4b9b775860539b63d7f6a75

suzuki-shunsuke:
Fixed. https://github.com/aquaproj/aqua/pull/1798/commits/aa100d4e16a6ac1a9be79dd3815c679de1784b31

```console
$ go run ./cmd/aqua gr 99designs/aws-vault
packages:
  - type: github_release
    repo_owner: 99designs
    repo_name: aws-vault
    description: A vault for securely storing and accessing AWS credentials in development environments
    asset: aws-vault-{{.OS}}-{{.Arch}}
    format: raw
    overrides:
      - goos: darwin
        format: dmg
        asset: aws-vault-{{.OS}}-{{.Arch}}.{{.Format}}
    supported_envs:
      - linux
      - darwin
      - windows/arm64
    checksum:
      type: github_release
      asset: SHA256SUMS
      file_format: regexp
      algorithm: sha256
      pattern:
        checksum: "^(\\b[A-Fa-f0-9]{64}\\b)"
        file: "^\\b[A-Fa-f0-9]{64}\\b\\s+(\\S+)$"
```

pkg/unarchive/dmg.go
```diff
@@ -109,31 +105,27 @@ func (unarchiver *dmgUnarchiver) Unarchive(ctx context.Context, fs afero.Fs, bod
...
+	if err := writeDmgFile(m, body, dest); err != nil {
+		return fmt.Errorf("write a dmg file: %w", err)
 	}
 
 	exe := exec.New()
 	tmpMountPoint := destDir + string(filepath.Separator) + "mount"
 	_, hdiutilDetach, err := exe.HdiutilAttach(ctx, dest, tmpMountPoint)
 	if err != nil {
-		return fmt.Errorf("hdiutil attach failed: %w", err)
+		return fmt.Errorf("hdiutil attach: %w", err)
```
suzuki-shunsuke, line 116: In this project, basically we don't include a word `failed` in `fmt.Errorf` because it is obvious the command failed and `failed` has no meaning.

pkg/unarchive/dmg.go
```diff
@@ -109,31 +105,27 @@ func (unarchiver *dmgUnarchiver) Unarchive(ctx context.Context, fs afero.Fs, bod
...
 	}
 
-	err = cpDir(fs, tmpMountPoint, destDir)
-	if err != nil {
-		return fmt.Errorf("failed cpDir src:%s,dst:%s :%w", tmpMountPoint, destDir, err)
+	if err := cpDir(fs, tmpMountPoint, destDir); err != nil {
```
suzuki-shunsuke, line 119: nits. I'd like to make the scope of the error variable small.

pkg/unarchive/dmg.go
```diff
@@ -0,0 +1,132 @@
...
+			if err := cpFile(fs, srcPath, dstPath); err != nil {
+				return fmt.Errorf("copy a file: %w", err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func writeDmgFile(m io.Writer, body io.Reader, dest string) error {
```
suzuki-shunsuke, line 69: Is this function necessary?
I think we can simply use `io.Copy` instead.

u-yas: When I tried purely using io.Copy, there was no particular problem...

I was writing the code by fumbling, so I didn't think about writing it more simply.

https://github.com/aquaproj/aqua/pull/1798/commits/d8546e9ab06f925b68c0c3f9bb566986005340b4

pkg/unarchive/dmg.go
u-yas: I agree! Thank you!

suzuki-shunsuke:
Added a workflow to test on macOS.
We run this workflow manually when we want to test on macOS.
Of course we can run this workflow by `pull_request` trigger, but I don't want to run this workflow everytime because macos runner is slower than linux. 4834e7cc461482d2d513244eaffe5fe9d78e59a4

pkg/exec/exec.go
```diff
@@ -65,14 +65,12 @@ func (exe *Executor) GoInstall(ctx context.Context, path, gobin string) (int, er
...
 }
 
-func hdiutilDetach(ctx context.Context, exe *Executor, mountPath string) (int, error) {
+func (exe *Executor) HdiutilDetach(ctx context.Context, mountPath string) (int, error) {
 	cmd := exe.command(exec.Command("hdiutil", "detach", mountPath, "-quiet"))
 	return exe.exec(ctx, cmd)
 }
 
-func (exe *Executor) HdiutilAttach(ctx context.Context, dmgPath, mountPoint string) (int, func(ctx context.Context, exe *Executor, mountPath string) (int, error), error) {
+func (exe *Executor) HdiutilAttach(ctx context.Context, dmgPath, mountPoint string) (int, error) {
```
suzuki-shunsuke, line 73: It was complicated to return a function, so I made this simpler.

pkg/unarchive/dmg.go
```diff
@@ -91,9 +96,8 @@ func (unarchiver *dmgUnarchiver) Unarchive(ctx context.Context, fs afero.Fs, bod
 		return fmt.Errorf("write a dmg file: %w", err)
 	}
 
-	exe := exec.New()
```
suzuki-shunsuke: Replace `exe` to the interface `unarchiver.executor` for testing.

pkg/unarchive/dmg.go
```diff
@@ -70,44 +70,40 @@ func cpDir(fs afero.Fs, src, dst string) error {
...
+	if _, err := unarchiver.executor.HdiutilAttach(ctx, tempFile.Name(), tmpMountPoint); err != nil {
 		return fmt.Errorf("hdiutil attach: %w", err)
 	}
 
-	if err := cpDir(fs, tmpMountPoint, destDir); err != nil {
+	if err := cpDir(fs, tmpMountPoint, unarchiver.dest); err != nil {
 		return fmt.Errorf("copy a directory: %w", err)
 	}
 
-	if err := fs.Remove(dest); err != nil {
```
suzuki-shunsuke: At first, I was confused because `dest` was removed.
`dest` was used as a temporal file but this was confusing.
So I used `fs.TempFile` instead.

pkg/exec/exec.go
```diff
@@ -64,3 +64,13 @@ func (exe *Executor) GoInstall(ctx context.Context, path, gobin string) (int, er
 	cmd.Env = append(os.Environ(), "GOBIN="+gobin)
 	return exe.exec(ctx, cmd)
 }
+
+func (exe *Executor) HdiutilDetach(ctx context.Context, mountPath string) (int, error) {
+	cmd := exe.command(exec.Command("hdiutil", "detach", mountPath, "-quiet"))
```
suzuki-shunsuke, line 69: I'm considering if we should set `-quiet` option or not.

`-quiet` option removes the error message.

```console
$ hdiutil attach  get.volta.sh
hdiutil: attach failed - image not recognized

$ hdiutil attach  get.volta.sh -quiet
$ echo $?
1
```

On the other hand, `-quiet` option makes the log clean.

pkg/exec/exec.go
```diff
@@ -64,3 +64,13 @@ func (exe *Executor) GoInstall(ctx context.Context, path, gobin string) (int, er
...
+	return exe.exec(ctx, cmd)
+}
+
+func (exe *Executor) HdiutilAttach(ctx context.Context, dmgPath, mountPoint string) (int, error) {
+	cmd := exe.command(exec.Command("hdiutil", "attach", dmgPath, "-mountpoint", mountPoint, "-quiet"))
```
suzuki-shunsuke, line 74: I'm considering if we should set `-quiet` option or not.

`-quiet` option removes the error message.

```console
$ hdiutil attach  get.volta.sh
hdiutil: attach failed - image not recognized

$ hdiutil attach  get.volta.sh -quiet
$ echo $?
1
```

On the other hand, `-quiet` option makes the log clean.

u-yas: How about removing the quiet option when AQUA_LOG_LEVEL=debug?

suzuki-shunsuke: Interesting.
Or it is also good to output logs only when the command failed.

u-yas: I don't think you can tell the success or failure of hdiutil at the stage of attaching the -quiet option.  If I run without the -quiet option, I feel that the hdiutil log should have flowed to the terminal without permission.  So I think that there is no choice but to control with the presence or absence of the log level in advance.  How about that?

suzuki-shunsuke: Done. cd205a7faa0d21026ae03bbb45e5750c9e5af129

suzuki-shunsuke: > How about removing the quiet option when AQUA_LOG_LEVEL=debug?

This idea is interesting, but I think if `hdiutil attach` command fails the error log should be always outputted.
Otherwise, users can't understand why `hdiutil attach` failed.

suzuki-shunsuke: > I don't think you can tell the success or failure of hdiutil at the stage of attaching the -quiet option. If I run without the -quiet option, I feel that the hdiutil log should have flowed to the terminal without permission. So I think that there is no choice but to control with the presence or absence of the log level in advance.

Sorry. I missed your comment.

I think we can judge if the command succeeded or failed as the following before outputting the log.
What do you think?

https://github.com/aquaproj/aqua/blob/cd205a7faa0d21026ae03bbb45e5750c9e5af129/pkg/exec/exec.go#L47-L58

suzuki-shunsuke:
I'd like to test this feature with more a few packages if possible.

pkg/unarchive/dmg.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func cpFile(fs afero.Fs, src, dst string) error {
+	srcFile, err := fs.Open(src)
+	if err != nil {
+		return fmt.Errorf("open a file: %w", err)
+	}
+	defer srcFile.Close()
+
+	dstFile, err := fs.Create(dst)
```
suzuki-shunsuke, line 33: Maybe we should use the third party library such as https://github.com/otiai10/copy instead.
The current implementation is simple and maybe there are issues.
For example, the current implementation doesn't keep the file attributes such as permissions.
Of course we would be able to fix issues ourselves, but using the third party library we don't have to maintain the code.

But unfortunately, https://github.com/otiai10/copy doesn't support afero.

suzuki-shunsuke: Done. 9d6d87fcce192d5396f48e1e418356e347cb364e

suzuki-shunsuke (approved):

u-yas:
@suzuki-shunsuke 

Thank you for your corrections. I tried to pull the latest commit and run it, but unfortunately I got the following error...

```
FATA[0001] aqua failed                                   aqua_version=2.1.0 env=darwin/arm64 error="get the unarchiver or decompressor by the file extension: get the unarchiver or decompressor by the file extension: format unrecognized by filename: .dmg" exe_name=aws-vault package=99designs/aws-vault package_version=v7.1.2 program=aqua
```

suzuki-shunsuke:
> aqua_version=2.1.0

Please run `go run ./cmd/aqua`

u-yas:
Ahhh

Maybe after doing go run ./cmd/aqua i and trying to run aws-vault, my locally installed brew installed aqua is running...

u-yas:
`go run ./cmd/aqua exec --aws-vault --version`
successfully executed! thank you

suzuki-shunsuke:
Looks good to me.

suzuki-shunsuke:
🎉 Released. https://github.com/aquaproj/aqua/releases/tag/v2.2.0

PR #1864, opened 2023-04-11
fix: fix the error handling of evaluating the version_constraint

suzuki-shunsuke:


pkg/config/registry/version_override.go
```diff
@@ -4,35 +4,39 @@ import (
...
 		if !strings.HasPrefix(v, prefix) {
-			return nil, nil //nolint:nilnil
+			return nil
 		}
 		sv = strings.TrimPrefix(v, prefix)
 	}
 	a, err := expr.EvaluateVersionConstraints(pkgInfo.VersionConstraints, v, sv)
 	if err != nil {
-		return nil, err //nolint:wrapcheck
+		logerr.WithError(logE, err).Debug("evaluate the version_constraint")
```
suzuki-shunsuke, line 22: If it fails to evaluate `version_constraint`, output debug log and the evaluation result is treated as `false`.

pkg/config/registry/version_override.go
```diff
@@ -48,7 +52,8 @@ func (pkgInfo *PackageInfo) SetVersion(v string) (*PackageInfo, error) {
 		}
 		a, err := expr.EvaluateVersionConstraints(vo.VersionConstraints, v, sv)
 		if err != nil {
-			return nil, err //nolint:wrapcheck
+			logerr.WithError(logE, err).Debug("evaluate the version_constraint")
```
suzuki-shunsuke, line 55: If it fails to evaluate `version_constraint`, output debug log and the evaluation result is treated as `false`.

pkg/policy/policy_package.go
```diff
@@ -61,14 +63,15 @@ type paramValidatePackage struct {
...
-func (pc *Checker) validatePackage(param *paramValidatePackage) error {
+func (pc *Checker) validatePackage(logE *logrus.Entry, param *paramValidatePackage) error {
 	if param.PolicyConfig == nil {
 		return nil
 	}
 	for _, policyPkg := range param.PolicyConfig.Packages {
 		f, err := pc.matchPkg(param.Pkg, policyPkg)
 		if err != nil {
-			return err
+			logerr.WithError(logE, err).Debug("check if the package matches with a policy")
```
suzuki-shunsuke, line 73: If it fails to check if the policy matches with the package, outputs debug log and treats as the policy doesn't match with the package.

PR #2089, opened 2023-07-10
fix: fix a bug that complete_windows_ext doesn't work

suzuki-shunsuke:


suzuki-shunsuke:
Duplicate of https://github.com/aquaproj/aqua/pull/2098

PR #2098, opened 2023-07-16
fix: fix a bug that file extensions aren't added when packages are installed on Windows

suzuki-shunsuke:
## AS IS

https://github.com/aquaproj/aqua/actions/runs/5570431513/jobs/10174468288

## TO BE

https://github.com/aquaproj/aqua/actions/runs/5570444206/jobs/10174492033


suzuki-shunsuke:
https://github.com/aquaproj/aqua/releases/tag/v2.9.1-0

PR #2413, opened 2023-10-30
feat: change the style of generate-registry

suzuki-shunsuke:


suzuki-shunsuke:
TODO

- Trim "v" from version_constraints
- Don't use `Version ==` in the latest version_constraitns
-  Fix a bug that the oldest version is duplicated

suzuki-shunsuke:
v2.16.5-2

PR #2553, opened 2023-12-06
chore: update golangci-lint

suzuki-shunsuke:


.golangci.yml
```diff
@@ -1,4 +1,8 @@
 ---
+run:
+  skip-files:
+    - "mock\\.go$"
+    - "mock_[^/]*\\.go$"
```
suzuki-shunsuke, line 5: Lint errors against mock is very bothersome.
So I exclude them.

PR #2626, opened 2024-01-24
docs: add CADDi to who-uses-aqua

akitok:
ref. https://github.com/orgs/aquaproj/discussions/2124


suzuki-shunsuke:
We're really glad that CADDi uses aqua!

PR #2929, opened 2024-05-31
feat: support fish completion

suzuki-shunsuke:


pkg/cli/completion.go
```diff
@@ -11,17 +11,27 @@ func (r *Runner) newCompletionCommand() *cli.Command {
...
-if command -v aqua &> /dev/null; then source <(aqua completion zsh); fi
+if command -v aqua &> /dev/null; then
+	source <(aqua completion zsh)
+fi
+
+~/.config/config.fish
+
+if command -v aqua &> /dev/null; 
+	aqua completion fish | source
+end
```
reitzig, line 34: FWIW, that does not seem to be the recommended way: https://fishshell.com/docs/current/completions.html#where-to-put-completions
(I used that command in the issue for brevity)

Rather, I would run

```bash
aqua completion fish > .config/fish/completions/aqua/fish
```

once after installation/upgrade. No additional hook is necessary.

suzuki-shunsuke: Thank you for your review.
Fixed. dee19a29b0349d7ff6b2af7c785bf5aaf5e093f9

PR #2960, opened 2024-06-18
feat: add `info package` subcommand

plobsing:
Provides one of the commands discussed in
https://github.com/orgs/aquaproj/discussions/2641

Output looks like

```
❯ aqua info package sigstore/cosign
{
  "type": "github_release",
  "repo_owner": "sigstore",
  "repo_name": "cosign",
  "description": "Container Signing",
  "asset": "cosign-{{.OS}}-{{.Arch}}",
  "format": "raw",
  "supported_envs": [
    "darwin",
    "linux",
    "amd64"
  ],
  "checksum": {
    "type": "github_release",
    "asset": "cosign_checksums.txt",
    "algorithm": "sha256"
  }
}
```

### DIFF
--- a/pkg/cli/package_info.go
+++ b/pkg/cli/package_info.go
@@ -1,0 +1,31 @@
+package cli //nolint:dupl
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/aquaproj/aqua/v2/pkg/config"
+	"github.com/aquaproj/aqua/v2/pkg/controller"
+	"github.com/urfave/cli/v2"
+)
+
+func (r *Runner) newPackageInfoCommand() *cli.Command {
+	return &cli.Command{
+		Name: "package",
+		Usage: "Show package defintions",
+		Description: `Show definition of packages.
+e.g.
+$ aqua info package sigstore/cosign`,
+		Action: r.packageInfoAction,
+	}
+}
+
+func (r *Runner) packageInfoAction(c *cli.Context) error {
+	param := &config.Param{}
+	if err := r.setParam(c, "generate", param); err != nil {
+		return fmt.Errorf("parse the command line arguments: %w", err)
+	}
+
+	ctrl := controller.InitializePackageInfoCommandController(c.Context, param, http.DefaultClient, r.Runtime)
+	return ctrl.PackageInfo(c.Context, param, r.LogE, c.Args().Slice()...)
+}
--- a/pkg/controller/packageinfo/controller.go
+++ b/pkg/controller/packageinfo/controller.go
@@ -1,0 +1,127 @@
+package packageinfo
+
+import (
+	"errors"
+	"encoding/json"
+	"strings"
+	"fmt"
+	"io"
+	"os"
+	"context"
+
+	"github.com/aquaproj/aqua/v2/pkg/config"
+	"github.com/aquaproj/aqua/v2/pkg/config/registry"
+	"github.com/spf13/afero"
+	"github.com/aquaproj/aqua/v2/pkg/checksum"
+	"github.com/aquaproj/aqua/v2/pkg/config/aqua"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	errMissingPackages = errors.New("could not find all packages")
+)
+
+type Controller struct {
+	stdout            io.Writer
+	configFinder      ConfigFinder
+	configReader      ConfigReader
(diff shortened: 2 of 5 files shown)


pkg/controller/packageinfo/controller.go
```diff
@@ -0,0 +1,127 @@
...
+
+	if missingPackages {
+		return errMissingPackages
+	}
+
+	return nil
+}
+
+func registryAndPackage(name string) (string, string) {
+	i := strings.IndexByte(name, ',')
```
suzuki-shunsuke, line 122: I think strings.Cut is better.

suzuki-shunsuke:

pkg/cli/package_info.go
```diff
@@ -0,0 +1,31 @@
...
+		Description: `Show definition of packages.
+e.g.
+$ aqua info package sigstore/cosign`,
+		Action: r.packageInfoAction,
+	}
+}
+
+func (r *Runner) packageInfoAction(c *cli.Context) error {
+	param := &config.Param{}
+	if err := r.setParam(c, "generate", param); err != nil {
```
suzuki-shunsuke, line 25:
```suggestion
	if err := r.setParam(c, "info-package", param); err != nil {
```

pkg/controller/packageinfo/controller.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return err
+	}
+
+	packageInfoByNameByRegistry := make(map[string]map[string]*registry.PackageInfo)
+	for registryName, registryConfig := range registryContents {
+		packageInfoByNameByRegistry[registryName]= registryConfig.PackageInfos.ToMap(logE)
+	}
+
+	encoder := json.NewEncoder(os.Stdout)
```
suzuki-shunsuke, line 93:
```suggestion
	encoder := json.NewEncoder(c.stdout)
```

pkg/controller/packageinfo/controller.go
```diff
@@ -0,0 +1,127 @@
...
+		if !ok {
+			logE.WithFields(logrus.Fields {
+				"registry_name": registryName,
+				"package_name": pkgName,
+			}).Error("no such registry or package")
+			missingPackages = true
+			continue
+		}
+
+		if err := encoder.Encode(pkgInfo); err != nil {
```
suzuki-shunsuke, line 109: Hmm. This command accepts multiple package names and outputs them as JSON independently.
I think it's weird because the output can't be parsed as JSON.

```json
{
  // ...
}
{
  // ...
}
```

I think there are several options.

1. Accept only one package
2. Return a list of packages

```json
[
  {
    // ...
  },
  {
    // ...
  }
]
```

pkg/controller/packageinfo/controller.go
```diff
@@ -0,0 +1,127 @@
...
+
+	return nil
+}
+
+func registryAndPackage(name string) (string, string) {
+	i := strings.IndexByte(name, ',')
+	if i == -1 {
+		return "standard", name
+	}
+	return name[:i], name[i+1:]
```
suzuki-shunsuke, line 126:
```suggestion
	r, p, ok := strings.Cut(name, ",")
	if ok {
		return r, p
	}
	return "standard", name
```

pkg/controller/packageinfo/controller.go
plobsing: JSON is a self-delimiting format. There is no parsing ambiguity in a stream of messages. Tools like `jq` and libraries such as Go's `encoding/json` have no issues working with such streams.

For example, with the current implementation, I'm able to aggregate a count of installed packages by owner using `jq`:
```
❮ ./aqua info package $(./aqua list --installed | awk '{print $3 "," $1}') | jq -r '.repo_owner' | sort | uniq -c 
   1 golangci
   1 goreleaser
   1 int128
   1 reviewdog
   1 rhysd
   1 sigstore
   2 suzuki-shunsuke
```

Of the alternatives you suggest, I think I prefer "accept only one package". Its got what seems to me like an unnecessary restriction, but the output is direct and simple to work with. It is a subset of the stream-of-objects output format I proposed: just pass a single name if you want a single message of output. If we do choose to restrict the caller to requesting only one package, what should happen if they pass more than one in `argv`?

I'd really prefer to avoid the list format. I find it over-encapsulated, less human-readable, and a bit less convenient to work with. There's not really an upside.

Issue #2996, opened 2024-07-12
Node.js support


W1M0R:
### Feature Overview

Please add node and npm to the package registry.

### Why is the feature needed?

Very convenient. If aqua supports node, then I don't need proto, pixi or vfox for node.

### Workaround

Use proto to install node.

### Note

I see the documentation clearly states that node is not supported: https://aquaproj.github.io/docs/#use-aqua-along-with-asdf

This is still a request to see what could be done about that limitation.

## Reference

- https://github.com/aquaproj/aqua-registry/pull/855
- https://github.com/aquaproj/aqua-registry/pull/1437
- https://github.com/aquaproj/aqua-registry/issues/1436

suzuki-shunsuke:
Please see this issue too.

- https://github.com/aquaproj/aqua-registry/issues/1436

suzuki-shunsuke:
Experimentally, I've added Node.js to aqua-registry again.

- https://github.com/aquaproj/aqua-registry/pull/26002
- https://github.com/aquaproj/aqua-registry/releases/tag/v4.214.0-1 (prerelease)

⚠️ For now, Windows isn't supported.

Please read the setup instruction in the release note too.

I'd really appreciate if you try this out and give us your feedback.

suzuki-shunsuke:
Oh, I found that the change of Node.js version isn't reflected...
We need to fix the setup instruction.

W1M0R:
Thanks @suzuki-shunsuke. I use aqua on Windows.

suzuki-shunsuke:
> I use aqua on Windows.

Oops. I see. I'm sorry about that.
I excluded Windows for now as [the content of zip file](https://nodejs.org/dist/v20.16.0/node-v20.16.0-win-x64.zip) was different from macOS and Linux, and it's a bit troublesome to verify on Windows.
After Linux and macOS, I'll work on Windows.

suzuki-shunsuke:
Windows has a different API of environment variables from Linux and macOS, so maybe it's difficult.
I'm not familiar with Windows.

W1M0R:
Here is how [Scoop](https://scoop.sh/) installs [Node](https://github.com/okibcn/ScoopMaster/blob/master/bucket/nodejs.json) on Windows:

```json
{
    "version": "22.6.0",
    "description": "An asynchronous event driven JavaScript runtime designed to build scalable network applications.",
    "homepage": "https://nodejs.org",
    "license": "MIT",
    "architecture": {
[...]

PR #3053, opened 2024-08-20
feat: add the subcommand `set-shell` to change the environment variable `PATH` for Node.js and Python Support

suzuki-shunsuke:
- https://github.com/aquaproj/aqua/issues/2996
- https://github.com/aquaproj/aqua-registry/pull/26002

Adding the subcommand `set-shell`, which outputs the script to change the environment variable `$PATH` dynamically.

```sh
. "$(aqua set-shell <shell type>)"
```

The support shell types are `zsh` and `bash`.

To support Node.js and Python, aqua needs to be able to change the environment variable `$PATH` dynamically because npm and pip install tools in the same directory.

### DIFF
--- a/.github/workflows/wc-integration-test.yaml
+++ b/.github/workflows/wc-integration-test.yaml
@@ -70,13 +70,14 @@ jobs:
           cd v20.16.0
           aqua i
           eval "$(aqua set-shell bash)"
+          eval "$(aqua output-shell)"
           node -v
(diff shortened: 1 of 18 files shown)


suzuki-shunsuke:
We decided not to merge this pull request for now.

- https://github.com/aquaproj/aqua/issues/2996#issuecomment-2307937252

Issue #3130, opened 2024-09-26
Change `$0` when executing tools


suzuki-shunsuke:
### Feature Overview

Enable us to change `$0` when executing tools like `exec`'s `-a` option.

```sh
exec -a assumego granted "$@" # Execute granted as assumego
```

### Why is the feature needed?

- https://github.com/aquaproj/aqua-registry/issues/27177#issuecomment-2377987560

Some tools change their behavior by `$0`.

For example, `granted` changes the behavior based on `args[0]`.

https://github.com/common-fate/granted/blob/e8de3ec7d62d543062d8be802b27abb3d8fac429/cmd/granted/main.go#L37-L44

```go
	// Use a single binary to keep keychain ACLs simple, swapping behavior via argv[0]
	var app *cli.App
	switch filepath.Base(os.Args[0]) {
	case "assumego", "assumego.exe", "dassumego", "dassumego.exe":
		app = assume.GetCliApp()
	default:
		app = granted.GetCliApp()
	}
```

### Workaround

Create shell scripts like this.

https://github.com/aquaproj/aqua-registry/issues/27177#issuecomment-2378009669

`~/bin/assumego`

```sh
#!/usr/bin/env bash

set -eu

granted=$(aqua which granted) # Get the absolute path of granted
exec -a assumego "$granted" "$@" # Execute granted changing args[0] to `assumego` by `exec -a assumego`
```

### Example Code

registry.yaml

```yaml
files:
  - name: assumego
    src: granted
    arg0: assumego
```

suzuki-shunsuke:
aqua executes commands by `exec.Run` and `unix.Exec`.

- https://pkg.go.dev/os/exec#CommandContext
- https://pkg.go.dev/os/exec#Cmd.Run
- https://pkg.go.dev/golang.org/x/sys/unix#Exec

https://github.com/aquaproj/aqua/blob/57c7332384f48ee2e56a144bdb9806c7faf8fdc6/pkg/osexec/exec.go#L24

https://github.com/aquaproj/aqua/blob/57c7332384f48ee2e56a144bdb9806c7faf8fdc6/pkg/osexec/xsys.go#L14

I thought we can change arg0 by changing `exec.Cmd.Args[0]` and `unix.Exec's argv[0]`, but it seems not to work.

```sh
#!/usr/bin/env bash

set -eu

echo "0: $0"
echo "@: $@"

echo foo
```

os/exec

```go
package main

import (
	"log"
	"os"
	"os/exec"
[...]

PR #3134, opened 2024-09-28
feat: support changing $0 by symlink

suzuki-shunsuke:
- https://github.com/aquaproj/aqua/issues/3130

```yaml
        files:
          - name: granted
          - name: assume
          - name: assumego
            src: granted
            link: assumego # link is the relative path from src to the symlink
```


suzuki-shunsuke:
https://github.com/aquaproj/aqua/releases/tag/v2.36.0-1

Issue #682, opened 2022-05-01
Support getting version from arbitrary URL for http packages


frezbo:
## Feature Overview

Golang publishes the release information [here](https://golang.org/dl/?mode=json) available as json
NodeJS publishes the release information [here](https://nodejs.org/dist/index.json) available as json

it would be nice if the `http` package support parsing the json using an expression and gathering the latest version information (this is sometimes required as the latest github release might not be the latest according to semver and could be just a backported release)

Just putting out here for ideas, currently I manage `go` and `nodejs` using a custom ansible code to get the latest versions which parse the release information from the json

Gowiem:
To reiterate what I said in https://github.com/aquaproj/aqua-registry/issues/14082, I think this could be valuable and is worth doing. Here was thoughts on schema to support this: 

> I could see some schema like this working:
> 
> ```yaml
> packages:
>   - name: snaplet/cli
>     type: http
>     package_version_fetch:
>        type: http
>        json_path: .content.versions.latest
>    ... 
> ```

suzuki-shunsuke:
- https://github.com/aquaproj/aqua/issues/3198

> ### Feature Overview
> 
> Currently there is no way to handle interactively (select, update, remove, etc.) non released "nightly" versions.
> 
> ### Why is the feature needed?
> 
> Maybe if aqua could support a new `json` value for `version_source` attribute, in order to declare how to fetch and parse (e.g. https://github.com/itchyny/gojq) the available versions, we would be able to declare how to fetch all "nightly" and "release" versions. And then work with those interactively via the CLI, like searching and selecting them, using `aqua g -s`.
> 
> ### Workaround
> 
> _No response_
> 
> ### Example Code
> 
> Configuration
> 
> e.g. for Zig:
> 
> ```yaml
> version_source: json
>   # - file: versions.json # or a local file
>   - url: https://ziglang.org/download/index.json
>     query: "(keys | del(.[] | select(. == "master"))), [.master.version] | .[]"
> ```
> 
> or for Node.js:
> 
> ```yaml
> version_source: json
>   - url: https://nodejs.org/download/release/index.json
>     query: "[.[].version]"
>   - url: https://nodejs.org/download/nightly/index.json
>     query: "[.[].version]"
> ```
> 
> ### Note
> 
> _No response_

[...]

PR #3269, opened 2024-11-19
feat: get versions from Go Proxy

suzuki-shunsuke:
This pull request adds a new field `go_version_path` to registry.

e.g.

```yaml
packages:
  - name: _go/sigsum.org/sigsum-go#cmd/sigsum-key
    type: go_install
    path: sigsum.org/sigsum-go/cmd/sigsum-key
    go_version_path: sigsum.org/sigsum-go
```

If this field is set, `aqua g` and `aqua up` commands gets available versions from [Go Module Proxy](https://proxy.golang.org/).

### DIFF
--- a/json-schema/registry.json
+++ b/json-schema/registry.json
@@ -359,6 +359,9 @@
         "path": {
           "type": "string"
         },
+        "go_version_path": {
+          "type": "string"
+        },
         "complete_windows_ext": {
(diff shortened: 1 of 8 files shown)


suzuki-shunsuke:
- https://github.com/aquaproj/aqua/issues/682

Issue #3294, opened 2024-11-27
Improvements on the example configuration for completion in Bash


akinomyoga:
### aqua info

```console
$ aqua info
{
  "version": "2.38.0",
  "commit_hash": "5ad46c267dda955e6355a5ee646f0d612dd9955b",
  "os": "linux",
  "arch": "amd64",
  "pwd": "/home/(USER)",
  "root_dir": "/home/(USER)/.local/share/aquaproj-aqua",
  "env": {},
  "config_files": []
}
```

I note that the above command and all the subsequent commands are executed in a Bash session.

### Overview

It's not really a bug of aqua, but I'd like to submit pull requests (PRs) to the example configuration of shell completions shown in the output of `aqua completion --help` and also on the [Usage](https://aquaproj.github.io/docs/reference/usage#aqua-completion) page in the online documentation.

I'm actually unsure about how to submit PRs to this project properly. I'm sorry if I'm doing it the wrong way. Since the contributing guide seems to require creating an issue before submitting a PR, let me here create one. There seem to be many specialized categories of the issue, but I somehow couldn't find one that seems to fit my case. I found a Slack link on the list of issue templates, so I thought of asking there how to choose the issue category, but I couldn't find out how to join the workspace. Then, although I still feel this is the wrong issue category, let me fill out the template as this is still a kind of fix.

I have three trivial fixes. Here, I primarily would like to consult with the maintainer about how I should organize my fixes in a proper granularity of PRs for this project.

### How to reproduce

aqua.yaml

I believe no config files are loaded as indicated in the output of `aqua info`. Since the output of `aqua completion --help` is anyway hardcoded in [the source code](https://github.com/aquaproj/aqua/blob/a43378144c4cf0c5077cc80e411b4de329d6480a/pkg/cli/completion/command.go#L19-L38), I think the output is reproducible regardless of the configuration.

```yaml

```

Other related code such as local Registry

I believe no other things are needed to reproduce the help text.

```

```

Executed command and output

```console
$ curl -sSfL https://raw.githubusercontent.com/aquaproj/aqua-installer/v3.0.1/aqua-installer | bash
$ PATH+=~/.local/share/aquaproj-aqua/bin
$ type -p aqua
/home/murase/.local/share/aquaproj-aqua/bin/aqua
$ aqua completion --help
NAME:
   aqua completion - Output shell completion script for bash, zsh, or fish

USAGE:
   aqua completion command [command options]

DESCRIPTION:
[...]

PR #3297, opened 2024-11-27
fix: improve the completion settings suggested in `aqua completion --help`

akinomyoga:
## Check List

  - Link to the issue: https://github.com/aquaproj/aqua/issues/3294

Three fixes are included in this PR.

## (1) `.bashrc` and `.zshrc` 4e21f1d09baa3f2e47efc47dfbb6617091777a9c

The current help text suggests adding the completion settings in `.bash_profile` and `.zprofile`, but `.bash_profile` and `.zprofile` are loaded only in login shell sessions. They are not loaded in (non-login) interactive sessions started inside the login shells, and thus the completion would be disabled in non-login interactive sessions.

**Actual**

```console
$ cat .bash_profile
if command -v aqua &> /dev/null; then
  source <(aqua completion bash)
fi
$ bash
$ complete -p aqua
<nothing is output>
```

**Expected**

```console
$ cat .bashrc
if command -v aqua &> /dev/null; then
  source <(aqua completion bash)
fi
$ bash
$ complete -p aqua
complete -o bashdefault -o default -o nospace -F _cli_bash_autocomplete aqua
```

In general, `.bash_profile` and `.zprofile` should only contain the settings that are automatically inherited by the child processes or the global user settings in the operating system.  The interactive settings should be put in `.bashrc` and `.zshrc`.

In this commit, we suggest `.bashrc` and `.zshrc` instead of `.bash_profile` and `.zprofile`, respectively.

## (2) Fix for Bash 3.* 8cbabba284a9cd5141561344a6f92e4303a88329

The currently suggested setting does not work for Bash 3.* because the `source` builtin in Bash 3.* does not support reading from a pipe and `source <(command)` fails silently.  Instead, `eval "$(command)"` works in all versions of Bash.

**Actual**

```console
$ bash-3.2
$ source <(aqua completion bash)
$ complete -p aqua
<nothing is output>
```

**Expected**

```console
$ bash-3.2
$ eval "$(aqua completion bash)"
$ complete -p aqua
complete -o bashdefault -o default -o nospace -F _cli_bash_autocomplete aqua
```

## (3) Style improvement 64ad3b42138c7f50fd4f137785a2ea412ac97cb2

This is an appearance improvement for the examples of different shells, and the change is just as you see in the diff.

### DIFF
--- a/pkg/cli/completion/command.go
+++ b/pkg/cli/completion/command.go
@@ -24,7 +24,7 @@ e.g.
 .bashrc
 
 if command -v aqua &> /dev/null; then
-	source <(aqua completion bash)
+	eval "$(aqua completion bash)"
 fi
 
 .zshrc
--- a/pkg/cli/completion/command.go
+++ b/pkg/cli/completion/command.go
@@ -21,13 +21,13 @@ Source the output to enable completion.
 
 e.g.
 
-.bash_profile
+.bashrc
 
 if command -v aqua &> /dev/null; then
 	source <(aqua completion bash)
 fi
 
-.zprofile
+.zshrc
 
 if command -v aqua &> /dev/null; then
 	source <(aqua completion zsh)
--- a/pkg/cli/completion/command.go
+++ b/pkg/cli/completion/command.go
@@ -21,19 +21,19 @@ Source the output to enable completion.
 
 e.g.
 
-.bashrc
+# .bashrc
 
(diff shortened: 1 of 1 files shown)


suzuki-shunsuke:

suzuki-shunsuke:
🎉 https://github.com/aquaproj/aqua/releases/tag/v2.38.1

PR #3362, opened 2024-12-15
Feat version expr

suzuki-shunsuke:


suzuki-shunsuke:
- https://github.com/aquaproj/aqua/pull/3363

PR #3442, opened 2025-01-07
feat(vacuum): add vacuum management for cleaning expired packages

NikitaCOEUR:
## Check List

- [X] Read [CONTRIBUTING.md](https://github.com/aquaproj/aqua/blob/main/CONTRIBUTING.md)
  - [X] Read [OSS Contribution Guide](https://github.com/suzuki-shunsuke/oss-contribution-guide/blob/main/README.md)
- [X] [Write a GitHub Issue before creating a Pull Request](https://github.com/suzuki-shunsuke/oss-contribution-guide/blob/main/README.md#create-an-issue-before-creating-a-pull-request)
  - Link to the issue:
- [X] [All commits are signed](https://github.com/suzuki-shunsuke/oss-contribution-guide/blob/main/docs/commit-signing.md)
  - This repository enables `Require signed commits`, so all commits must be signed
- [X] [Avoid force push](https://github.com/suzuki-shunsuke/oss-contribution-guide?tab=readme-ov-file#dont-do-force-pushes-after-opening-pull-requests)
- [] Do only one thing in one Pull Request

# Introduce Vacuum Feature for Expired Packages

Addresses [#2942](https://github.com/aquaproj/aqua/issues/2942) and [Discussion #3086](https://github.com/orgs/aquaproj/discussions/3086).

This pull request introduces a new feature for vacuuming expired packages.

## Workflow

1. **Environment Variable Requirement**  
   The feature activates only when the environment variable `AQUA_VACUUM_DAYS` is set with a positive integer greater than 0.

2. **Implementation in `InstallPackage`**  
   - During the execution of commands through `exec` or `install` (but not `aqua update` or `aqua cp`, which currently use `provideNilVacuumController` in their controller initialization), package-related information is added or updated in a BoltDB (`vacuum.db`) located in `ROOT_DIR`.  
   - The stored information includes:  
     - Package Type  
     - Package Name  
     - Version  
     - Timestamp (LastUsage)  
     - PkgPaths  

3. **Expiration Criteria**  
   - Packages that are not used do not have their timestamp updated.  
   - A package is considered expired if its last usage exceeds the number of days defined in `AQUA_VACUUM_DAYS`. Expired packages become eligible for "vacuum."

4. **Execution of `aqua vacuum` Command**  
   - The `vacuum` command performs the following operations:  
     - Deletes the package and its version from `ROOT_DIR/pkgs/type.../pkg/version`.  
     - Removes the corresponding entry from the BoltDB.

[...]

PR #3467, opened 2025-01-22
feat: vacuum unused packages

### LINKED ISSUE #2942: Command to remove outdated packages/binaries
### Feature Overview

As a (power 😇) user, I want to maintain my user-space, cross-project tooling with aqua without blowing up storage usage over time.

I understand that an automatism is hard to conceive of, due to the highly dynamic and semi-stateless way aqua works. However, a command for me to call would solve my use case; I would just integrate it into my upgrade routine.

When I frequently upgrade tools installed through aqua, it assembles quite a volume of files. After just a few days of usage, I have:

```console
🐧❯ du -sh ~/.local/share/aquaproj-aqua/
1,2G	/home/raphael/.local/share/aquaproj-aqua/
```

Meanwhile, a clean install results in

```console
🐧❯ aqua list -a --installed | wc -l
21
🐧❯ du -sh ~/.local/share/aquaproj-aqua
862M	/home/raphael/.local/share/aquaproj-aqua
```

With a few rather active tools in my list, I expect to need a cleanup routine sooner rather than later.

### Workaround

While I can definitely

```console
aqua rm --all
```

and rely on lazy install, depending on _when/where_ the install triggers, the experience is less than pretty. For instance, with [direnv/direnv](https://github.com/direnv/direnv) and [starship/starship](https://github.com/starship/starship):

```console
🐧❯ aqua rm --all
INFO[0000] removing all packages                         aqua_version=2.29.0 env=linux/amd64 program=aqua
/home/raphael/.local/share/aquaproj-aqua/pkgs/github_release/github.com/direnv/direnv/v2.34.0/direnv.linux-amd64/direnv.linux-amd64: command not found
- (line 3): 
        "/home/raphael/.local/share/aquaproj-aqua/pkgs/github_release/github.com/direnv/direnv/v2.34.0/direnv.linux-amd64/direnv.linux-amd64" export fish | source;
        ^
in function '__direnv_export_eval'
in event handler: handler for generic event “fish_prompt”
INFO[0000] download and unarchive the package            aqua_version=2.29.0 env=linux/amd64 exe_name=starship package_name=starship/starship package_version=v1.19.0 program=aqua registry=standard
[WARN] - (starship::modules::custom): Executing custom command "env | grep -c KUBE" timed out.
[WARN] - (starship::modules::custom): You can set command_timeout in your config to a higher value or set ignore_timeout to true for this module to allow longer-running commands to keep executing.
```

I guess I could do this as an "atomic" action:

```console
aqua rm --all && aqua install -a
```

But then I would always re-download _all_ tools, which seems a waste of time and bandwidth.

### Example Code

I imagine we could have something like:

```console
$ aqua rm --outdated
```

This would delete all versions except the latest ones.

(This would, of course, cause re-installs for all projects that pin an older version. For me, that'd be acceptable.)

Note sure how one would encode "delete all versions but the ones pinned in the global config" -- `--all` means different things to `aqua install` and `aqua rm`. 🤔 

Maybe a dedicated command like

```console
$ aqua clean [--all]
```

could delete all versions but the once references in the current scope?

### Note

I found this, which seems to leave room for future discussions:

> You may want to keep the latest version and delete only old versions, but we don't support it now.
Lazy Install will reinstall the latest version.
#538

So here I am. 😁

suzuki-shunsuke:
Close #2942
- https://github.com/orgs/aquaproj/discussions/3086

```sh
aqua vacuum [--days (-d) 60]
aqua vacuum --init
```

`aqua vacuum` removes packages and timestamp files if timestamp is older than vacuum days.

`aqua vacuum --init` searches packages from aqua.yaml including global configuration files `$AQUA_GLOBAL_CONFIG` and creates timestamp files if they are installed.

## How does it work?

aqua manages each package's last used datetime in `$(aqua root-dir)/metadata/<package path>/timestamp.txt`.
aqua creates or updates these files when packages are installed or executed.
`aqua vacuum` removes packages and timestamp files if timestamp is older than vacuum days.

Until packages are installed or executed, timestamp files aren't created.
If packages don't have timestamp files, `aqua vacuum` doesn't remove those packages.

To solve the problem, `aqua vacuum --init` searches packages from aqua.yaml including global configuration files `$AQUA_GLOBAL_CONFIG` and creates timestamp files if they are installed.

## Compared with #3442

#3442 uses boltDB to manage last used dates.
On the other hand, this pr uses text files to manage them.
This pull request is much simpler than #3442 and the overhead is much cheaper.
We don't need to learn boltDB, and we don't need to maintain complicated asynchronous codes.
Complicated code makes the maintenance hard and raises bugs easily.
So simplicity is justice.

This pull request doesn't implement functions to show timestamps because the feature is used only for debug.
This pull request implements the function to initialize timestamps by the current datetime.

## Performance

```console
$ AQUA_VACUUM_DAYS=5 hyperfine -r 100 -N --warmup 3 '/Users/shunsukesuzuki/go/bin/aqua exec -- cmdx -v' '/Users/shunsukesuzuki/.local/share/aquaproj-aqua/internal/pkgs/github_release/github.com/aquaproj/aqua/v2.42.2/aqua_darwin_arm64.tar.gz/aqua exec -- cmdx -v'

Benchmark 1: /Users/shunsukesuzuki/go/bin/aqua exec -- cmdx -v
  Time (mean ± σ):      38.7 ms ±   0.9 ms    [User: 6.4 ms, System: 1.9 ms]
  Range (min … max):    37.3 ms …  42.1 ms    100 runs
 
Benchmark 2: /Users/shunsukesuzuki/.local/share/aquaproj-aqua/internal/pkgs/github_release/github.com/aquaproj/aqua/v2.42.2/aqua_darwin_arm64.tar.gz/aqua exec -- cmdx -v
  Time (mean ± σ):      41.0 ms ±   1.3 ms    [User: 5.7 ms, System: 1.8 ms]
  Range (min … max):    39.0 ms …  44.3 ms    100 runs
 
Summary
  /Users/shunsukesuzuki/go/bin/aqua exec -- cmdx -v ran
    1.06 ± 0.04 times faster than /Users/shunsukesuzuki/.local/share/aquaproj-aqua/internal/pkgs/github_release/github.com/aquaproj/aqua/v2.42.2/aqua_darwin_arm64.tar.gz/aqua exec -- cmdx -v
```

### DIFF
--- a/.github/workflows/wc-integration-test.yaml
+++ b/.github/workflows/wc-integration-test.yaml
@@ -203,6 +203,19 @@ jobs:
           git checkout -- .
         working-directory: tests/update3
 
+      - name: Test aqua install with AQUA_VACUUM_DAYS
+        env:
+          AQUA_VACUUM_DAYS: 1
+        run: aqua install
+      - name: Test aqua exec with AQUA_VACUUM_DAYS
+        env:
+          AQUA_VACUUM_DAYS: 1
+        run: cmdx -v
+      - name: Test vacuum Run command
+        env:
+          AQUA_VACUUM_DAYS: 1
+        run: aqua vacuum
+
       - run: aqua update-checksum -a
 
       - run: terraform --help
--- a/.github/workflows/wc-integration-test.yaml
+++ b/.github/workflows/wc-integration-test.yaml
@@ -203,18 +203,12 @@ jobs:
           git checkout -- .
         working-directory: tests/update3
 
-      - name: Test aqua install with AQUA_VACUUM_DAYS
-        env:
-          AQUA_VACUUM_DAYS: 1
-        run: aqua install
-      - name: Test aqua exec with AQUA_VACUUM_DAYS
-        env:
-          AQUA_VACUUM_DAYS: 1
-        run: cmdx -v
-      - name: Test vacuum Run command
+      - name: Test vacuum
         env:
           AQUA_VACUUM_DAYS: 1
         run: aqua vacuum
+      - name: Test vacuum --init
+        run: aqua vacuum --init
 
       - run: aqua update-checksum -a
 
--- a/pkg/cli/runner.go
+++ b/pkg/cli/runner.go
@@ -20,6 +20,7 @@ import (
 	"github.com/aquaproj/aqua/v2/pkg/cli/update"
 	"github.com/aquaproj/aqua/v2/pkg/cli/updateaqua"
 	"github.com/aquaproj/aqua/v2/pkg/cli/util"
+	"github.com/aquaproj/aqua/v2/pkg/cli/vacuum"
 	"github.com/aquaproj/aqua/v2/pkg/cli/version"
 	"github.com/aquaproj/aqua/v2/pkg/cli/which"
 	"github.com/urfave/cli/v2"
@@ -110,6 +111,7 @@ func Run(ctx context.Context, param *util.Param, args ...string) error { //nolin
 			upc.New,
 			remove.New,
 			update.New,
+			vacuum.New,
 		),
 	}
 
--- a/pkg/cli/runner.go
+++ b/pkg/cli/runner.go
@@ -84,12 +84,6 @@ func Run(ctx context.Context, param *util.Param, args ...string) error { //nolin
 				Name:  "cpu-profile",
 				Usage: "cpu profile output file path",
 			},
-			&cli.IntFlag{
-				Name:    "vacuum-days",
-				Usage:   "Vacuum days",
-				EnvVars: []string{"AQUA_VACUUM_DAYS"},
-				Value:   60, //nolint:mnd
-			},
 		},
 		EnableBashCompletion: true,
 		Commands: commands(
--- a/pkg/cli/runner.go
+++ b/pkg/cli/runner.go
@@ -83,6 +83,11 @@ func Run(ctx context.Context, param *util.Param, args ...string) error { //nolin
 				Name:  "cpu-profile",
 				Usage: "cpu profile output file path",
 			},
+			&cli.IntFlag{
+				Name:    "vacuum-days",
+				Usage:   "Vacuum days",
+				EnvVars: []string{"AQUA_VACUUM_DAYS"},
+			},
 		},
 		EnableBashCompletion: true,
 		Commands: commands(
--- a/pkg/cli/runner.go
+++ b/pkg/cli/runner.go
@@ -88,6 +88,7 @@ func Run(ctx context.Context, param *util.Param, args ...string) error { //nolin
 				Name:    "vacuum-days",
 				Usage:   "Vacuum days",
 				EnvVars: []string{"AQUA_VACUUM_DAYS"},
+				Value:   60, //nolint:mnd
 			},
 		},
 		EnableBashCompletion: true,
--- a/pkg/cli/util/util.go
+++ b/pkg/cli/util/util.go
@@ -137,9 +137,9 @@ func SetParam(c *cli.Context, logE *logrus.Entry, commandName string, param *con
 		}
 		param.EnforceRequireChecksum = requireChecksum
 	}
-	param.VacuumDays = c.Int("vacuum-days")
+	param.VacuumDays = c.Int("days")
 	if param.VacuumDays <= 0 {
-		return errors.New("vacuum-days must be greater than 0")
+		return errors.New("vacuum days must be greater than 0")
 	}
 	return nil
 }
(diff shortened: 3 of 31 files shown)


suzuki-shunsuke:
# Is the command name `vacuum` good?

There are several options

- Rename `vacuum` to `clean` or something
- Merge `vacuum` with `remove` command

NikitaCOEUR:
## About command name
> # Is the command name `vacuum` good?
> 
> There are several options
> 
>     * Rename `vacuum` to `clean` or something
> 
>     * Merge `vacuum` with `remove` command

When I was working on the integration, I wanted to use "remove" but there were dependencies in the controller with elements that were, with my level in golang..., too important. It seemed to me that there was a dependency with the `aqua.yaml` files that I didn't want to be found again when I ran the "vacuum", deleting packages shouldn't require the configuration of a file like aqua.yaml.

If I have X repositories with different yaml configurations, I didn't want the command to interact only with the packages defined in one of them.

So I wanted to have a complete decorrelation with remove. And when I proposed the idea in the https://github.com/orgs/aquaproj/discussions/3086 discussion, I had thought of vacuum rather than clean, as I felt that clean was too close to remove in connotation. And I thought of Postgres, which implements the vacuum command. In the end, it seemed more appropriate.

Also, adding timestamp management to the use/download of each package seemed too far removed from the term "remove", I felt that adding a new feature responsible for the overall functionality of remembering the elements it needs to clean was more coherent.
So, in my opinion:
1. Add vacuum command.
2. Full integration with the remove command (with a vacuum parameter to execute this action).

What don't you like about "vacuum"?

## Whether or not to use a db bbolt
I would have loved to have my name on this repository code, I found the use of a bbolt really sexy.
However, aqua is a tool I appreciate, and you're someone I follow on the various projects you do. I'm really impressed by your presence, your ideas, your follow-through and your ability to think on your feet.
I know that you're (almost) the only one working on these projects, so it's vital that you master the tool and the code that make it up to perfection, and that you have the final say on how it's done.

But I still think that using a bbolt is more appropriate, as it seems less "cobbled-together". But the desire to limit overhead as much as possible forces the use of tricks and tends towards code complexity, as can be seen in https://github.com/aquaproj/aqua/pull/3442. (Asynchronous execution, even execution of an independent process to get around the limitation of exec commands in syscall_unix).

In short, you have the last word!

suzuki-shunsuke:
Thank you for your comment.

Now I think adding a vacuum command is best.

About the command name, note that I'm not a native English speaker and I'm not so good at English.
So my feeling may be incorrect.

The word "vacuum" brought to mind the noun meaning "a void" or "emptiness."
But seems like "vacuum" has a meaning of verb `to clean using a vacuum cleaner`, which I feel suitable.

https://www.ldoceonline.com/dictionary/vacuum

> I had thought of vacuum rather than clean, as I felt that clean was too close to remove in connotation.

Agree. `clean` is confusing.

About merging, I considered API.

e.g.

```sh
remove --expired [--init-timestamp, --init-vacuum] [--expired-days 60]
```

But I don't think this is user-friendly.
Both `vacuum` and `remove` are commands to remove packages, but the mechanism is totally different.
So now I think adding a new command is better for both users and maintainers rather than merging these commands.

> I would have loved to have my name on this repository code and i found the use of a bbolt really sexy.

Yeah, I'm so sorry to create a new pull request instead of your pull request.
I really appreciate and respect your effort, and I wanted to merge you pull request.
Without your effort, I would never work on this feature and create this pull request.

I think BoltDB is an awesome project. I was glad to learn it in this time.
Maybe I'll use it in my OSS or work.
But in this time I feel it's unnecessary.
If we will need it in future, we'll be able to reconsider it, but it's not now.

---

Ideally, I'd like to store metadata in the same directory with each package.

Now:

```
pkgs/<package path>/<command>
metadata/<package path>/timestamp.txt # Maybe we will add other files like metadata.json
```

Ideal:

(There is a room to improve file and directory names and structure)

```
pkgs/<package path>/
  aqua-package.json # A file to identify the package directory. This is useful to list installed packages. We can initialize all packages' timestamps using the current date time.
  metadata/
    timestamp.txt
  package/
    <command>
```

But this is a breaking change, so we gave up it.

suzuki-shunsuke:
- https://github.com/aquaproj/aquaproj.github.io/pull/1342

suzuki-shunsuke:
v2.43.0-0 is out.

You can try this version:

```sh
aqua upa v2.43.0-0
```

The document is here. https://github.com/aquaproj/aquaproj.github.io/pull/1342
I'll try this for a while (a few days?).

Note:

1. `AQUA_VACUUM_DAYS` is optional. Last used days are recorded when packages are installed or executed even if `AQUA_VACUUM_DAYS` is not set.
The default expiration days is 60.
1. I removed the command alias `v`. Perhaps we'll recover the alias if necessary

suzuki-shunsuke:
- `aqua rm --all` should remove `metadata` directory
- `aqua rm` should remove removed packages' `timestamp.txt`

PR #3649, opened 2025-03-14
feat: generate checksum cosign config

scop:
This adds support for generating cosign configs for checksum files in `gr`.

Lightly tested, may have bugs. But a happy path exists: https://github.com/aquaproj/aqua-registry/pull/33234

## Check List

  - Link to the issue: https://github.com/aquaproj/aqua-registry/issues/32207#issuecomment-2661324286
- [ ] Do only one thing in one Pull Request

### DIFF
--- a/pkg/controller/generate-registry/generate.go
+++ b/pkg/controller/generate-registry/generate.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/aquaproj/aqua/v2/pkg/asset"
@@ -244,6 +245,7 @@ func (c *Controller) patchRelease(logE *logrus.Entry, pkgInfo *registry.PackageI
 			if chksum != nil {
 				assetInfo := asset.ParseAssetName(checksumName, tagName)
 				chksum.Asset = assetInfo.Template
+				chksum.Cosign = checkChecksumCosign(pkgInfo, checksumName, assetNames)
 				pkgInfo.Checksum = chksum
 				break
 			}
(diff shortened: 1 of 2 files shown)


pkg/controller/generate-registry/group.go
```diff
@@ -248,7 +248,10 @@ func (c *Controller) group(logE *logrus.Entry, pkgName string, releases []*Relea
 
 	for _, group := range groups {
 		release := group.releases[0]
-		pkgInfo := &registry.PackageInfo{}
+		pkgInfo := &registry.PackageInfo{
+			RepoOwner: pkgInfo.RepoOwner,
+			RepoName:  pkgInfo.RepoName,
+		}
```
scop, line 254: This and carrying the pkgInfo around is not too nice at all, but I needed some way to get the repo owner and name down to patchRelease. Figured I'd start simple, this appears to work at least for some cases.

pkg/controller/generate-registry/generate.go
```diff
@@ -285,3 +287,45 @@ func checkSLSAProvenance(assetName, tagName string) *registry.SLSAProvenance {
...
+		return nil
+	}
+
+	downloadURL := fmt.Sprintf("https://github.com/%s/%s/releases/download/{{.Version}}/",
+		pkgInfo.RepoOwner, pkgInfo.RepoName)
+	return &registry.Cosign{
+		Opts: []string{
+			"--certificate-identity-regexp",
+			fmt.Sprintf(
+				`^https://github\.com/%s/%s/\.github/workflows/.+\.ya?ml@refs/tags/`,
```
scop, line 319: I left `{{.Version}}` out from here, mostly because I feel simply having the version replaced here literally by templating might not be a good thing to do as the version could contain regex metacharacters.

We could go with this though, on a second look I think I'd actually like this better. Thoughts?
```suggestion
				`^https://github\.com/%s/%s/\.github/workflows/.+\.ya?ml@refs/tags/\Q{{.Version}\E$`,
```

suzuki-shunsuke:

pkg/controller/generate-registry/generate.go
```diff
@@ -285,3 +287,45 @@ func checkSLSAProvenance(assetName, tagName string) *registry.SLSAProvenance {
...
+				regexp.QuoteMeta(pkgInfo.RepoOwner),
+				regexp.QuoteMeta(pkgInfo.RepoName),
```
suzuki-shunsuke, line 321:
```suggestion
				pkgInfo.RepoOwner,
				pkgInfo.RepoName,
```

In my understanding, regular expression metacharacters aren't available in GitHub repository owners and names.
So we don't need to use [regexp.QuoteMeta](https://pkg.go.dev/regexp#QuoteMeta) here.
But there is no problem even if it is used.

suzuki-shunsuke:
🎉 https://github.com/aquaproj/aqua/releases/tag/v2.46.0

PR #3665, opened 2025-03-16
feat: make trimAssetExt available in templates

scop:
For example for use in cosign options. Another alternative is to make AssetWithoutExt more widely available, but this would sound like a more general way to handle it.

## Check List

- [ ] Read [CONTRIBUTING.md](https://github.com/aquaproj/aqua/blob/main/CONTRIBUTING.md)
  - [ ] Read [OSS Contribution Guide](https://github.com/suzuki-shunsuke/oss-contribution-guide/blob/main/README.md)
- [ ] [Write a GitHub Issue before creating a Pull Request](https://github.com/suzuki-shunsuke/oss-contribution-guide/blob/main/README.md#create-an-issue-before-creating-a-pull-request)
  - Link to the issue:
- [ ] [All commits are signed](https://github.com/suzuki-shunsuke/oss-contribution-guide/blob/main/docs/commit-signing.md)
  - This repository enables `Require signed commits`, so all commits must be signed
- [ ] [Avoid force push](https://github.com/suzuki-shunsuke/oss-contribution-guide?tab=readme-ov-file#dont-do-force-pushes-after-opening-pull-requests)
- [ ] Do only one thing in one Pull Request

PR #3675, opened 2025-03-20
fix: set AssetWithoutExt

suzuki-shunsuke:
- https://github.com/aquaproj/aqua/pull/3665#issuecomment-2729260911


suzuki-shunsuke:
v2.46.0-2

PR #3722, opened 2025-04-04
feat: support slsa-verifier's -build-workflow-input

suzuki-shunsuke:


suzuki-shunsuke:
# ⚠️ Blocker

Probably we want to use template variables in input variables.
But the needs is not clear.
We shouldn't implement the feature until the needs becomes clear.

PR #3772, opened 2025-04-17
fix: fix `-v` option and version command

suzuki-shunsuke:


sheldonhull:
@suzuki-shunsuke this might mean another bootstrap installer to maintain compatibility fyi. 
I just had my pipelines fail due to not working with version. The v2.48.2 will contain the fix I'm assuming. The bootstrap is on older version. 

So if you bootstrap, run `aqua version` it will fail. This forces you to run bootstrap, run update, then output the version. That's fine, but I was providing the before and after versions to show the update. Just an FYI.

suzuki-shunsuke:
Thank you!
I could reproduce the issue.
I'm taking a look.

- https://github.com/aquaproj/aqua/issues/3773

PR #4006, opened 2025-07-16
fix: replace *http.DefaultClient with *http.Client for GitHub

### LINKED ISSUE #4005: Intermittent 403 errors when downloading private GitHub release
### aqua info

```console
$ aqua info
{
  "version": "2.53.4",
  "commit_hash": "Homebrew",
  "os": "darwin",
  "arch": "arm64",
  "pwd": "/Users/(USER)",
  "root_dir": "/Users/(USER)/.local/share/aquaproj-aqua",
  "env": {
    "GITHUB_TOKEN": "(masked)"
  },
  "config_files": []
}
```

### Overview

When running Aqua in GitHub Actions through the aqua-installer action, we occasionally get 403 responses while it downloads GitHub releases from our private repositories.

This started somewhere last week. We've been using Aqua without problem for several months so far. No changes were made on our side; this just started popping up suddenly last week.

All GitHub authentication works fine—if not, it would fail all the time instead of intermittently. In some projects, we download ~3 private packages, and sometimes, only 1 fails while the others succeed.

Whenever it fails, it shares what seems like an S3 signed URL:

```level=error msg=install the package env=linux/amd64 error=download the release asset (asset id: 243231222): GET https://release-assets.githubusercontent.com/github-production-release-asset/885343073/da0d91e3-daf6-4edd-a4fd-4153d3b3d108?sp=r&sv=2018-11-09&sr=b&spr=https&se=2025-07-15T16%3A11%3A54Z&[...]jwt=*** 403  []```

Hoping to fix this, we updated to the latest version of Aqua (both the CLI and the installer), but the problem persists.

### How to reproduce

Reproducing is very tricky since it's intermittent, but we get this issue with releases from private GitHub repositories.

### Expected behaviour

It should download the package.

### Actual behaviour

It returns a 403.

### Note

_No response_

suzuki-shunsuke:
Close #4005

Use authenticated http client when downloading assets using redirect URL.
Now aqua uses http.DefaultClient, so it doesn't retry downloading assets when HTTP 5xx error, and it doesn't have a GitHub access token, so it can't download assets from private repositories.

- Retry HTTP 5xx error
- Pass a GitHub Access token

### DIFF
--- a/pkg/download/github_release.go
+++ b/pkg/download/github_release.go
@@ -60,7 +60,7 @@ func (dl *GitHubReleaseDownloader) DownloadGitHubRelease(ctx context.Context, lo
 	if err != nil {
 		return nil, 0, err
 	}
-	body, redirectURL, err := dl.github.DownloadReleaseAsset(ctx, param.RepoOwner, param.RepoName, assetID, github.HTTPClient())
+	body, redirectURL, err := dl.github.DownloadReleaseAsset(ctx, param.RepoOwner, param.RepoName, assetID, github.GetHTTPClient(ctx, logE))
 	if err != nil {
 		return nil, 0, fmt.Errorf("download the release asset (asset id: %d): %w", assetID, err)
 	}
--- a/pkg/download/github_release.go
+++ b/pkg/download/github_release.go
@@ -60,7 +60,7 @@ func (dl *GitHubReleaseDownloader) DownloadGitHubRelease(ctx context.Context, lo
 	if err != nil {
 		return nil, 0, err
 	}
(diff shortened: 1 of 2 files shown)


SanderKnape:
I asked my AI friend and - of course, take this with a grain of salt - it suggests the following:

`The fix in the branch will not work unless the initialization order is changed so that github.New is called before any use of github.HTTPClient().`

suzuki-shunsuke:
I think github.New is called before `DownloadGitHubRelease`.

suzuki-shunsuke:
- https://github.com/aquaproj/aqua/pull/4019

Issue #4068, opened 2025-08-08
Checksum verification fails for aqua-registry (Failed to download correct registries from raw.githubusercontent.com)


t3-aoki:
### package name

The package is not a specific one; it is aqua-registry itself.

### aqua info

This happens in multiple versions, but here's an example:

```console
$ aqua info
{
  "version": "2.28.0",
  "commit_hash": "ecd9e704110bf2f236b744cd6497ad50e1b52e47",
  "os": "linux",
  "arch": "amd64",
  "pwd": "/workspaces/repo",
  "root_dir": "/home/(USER)/.local/share/aquaproj-aqua",
  "env": {
    "AQUA_GLOBAL_CONFIG": "/workspaces/repo/aqua.yaml"
  },
  "config_files": [
    {
      "path": "/workspaces/repo/aqua.yaml"
    }
  ]
}
```

### Overview

Since around 4:00 UTC on August 8, 2025, checksum verification for aqua-registry has started failing.
I use multiple versions of aqua and aqua-registry across several repositories, and the verification fails in all cases.

Examples error message:

```
time="2025-08-08T04:19:47Z" level=error msg="install the registry" actual_checksum=AA73FBEE71880E59A8D00E5A4AB869A8CAE8BD5BDA335B3959B12D35E8A1C9AA aqua_version=2.27.3 env=linux/amd64 error="check a registry's checksum: checksum is invalid" expected_checksum=16B8DC8967FBEC6F7E7B9205393DE8A06A49438373B71767FB73753F137D04DE program=aqua registry_name=standard
time="2025-08-08T04:19:47Z" level=fatal msg="aqua failed" aqua_version=2.27.3 env=linux/amd64 error="it failed to install some registries" program=aqua
```

```
time=2025-08-08T05:10:32Z level=error msg=install the registry actual_checksum=A74D64E68D2ED8328080DE582F8B563DE248143AB7B9EF5DDD671C5B2A626E256D650B7C787ED1D5ACB99634A63D2B0742817FF2EC9FFB430206D871212AEE7E env=linux/amd64 error=check a registry's checksum: checksum is invalid expected_checksum=D1DF243740112760AA0F76C8EB71B124BB6E4D2BF75F352322252E8A53C8ED2CC4A6A9A533C9281A42CF019046EF7BE91932C9590AD765FF997A258F1A8DA222 program=aqua program_version=2.53.9 registry_name=standard
[...]

PR #4069, opened 2025-08-08
fix: get a full length commit sha when downloading a file from GitHub Content

suzuki-shunsuke:


suzuki-shunsuke:
Probably the problem was solved on the GitHub side.

- https://github.com/aquaproj/aqua/issues/4068#issuecomment-3167583988

So I don't think this pr is necessary anymore.

PR #4342, opened 2025-11-18
perf(which): Skip packages that cannot provide the desired exe

refi64:
When looking for the package providing an exe, several time-consuming tasks need to be performed on each pkgInfo instance (in particular, evaluating the version constraints). But in most cases, the exe in question is not available across *any* version constraints or the like, in which case we can save some performance by skipping it entirely.

In the case where many packages have to be checked, this can give a sizable perf boost. Here I'm using `jq`, which is **not** installed by aqua (and thus every local *and* global package) has to be checked:

```
$ hyperfine --warmup=20 -N 'dist/aqua'{.old,}' which jq'
Benchmark 1: dist/aqua.old which jq
  Time (mean ± σ):       6.1 ms ±   0.7 ms    [User: 5.3 ms, System: 2.1 ms]
  Range (min … max):     4.6 ms …   9.8 ms    573 runs

Benchmark 2: dist/aqua which jq
  Time (mean ± σ):       4.8 ms ±   0.7 ms    [User: 3.9 ms, System: 1.8 ms]
  Range (min … max):     3.3 ms …   9.0 ms    610 runs

Summary
  dist/aqua which jq ran
    1.27 ± 0.24 times faster than dist/aqua.old which jq

$ hyperfine --warmup=20 -N 'dist/aqua'{.old,}' exec -- jq'
Benchmark 1: dist/aqua.old exec -- jq
  Time (mean ± σ):       7.8 ms ±   0.9 ms    [User: 6.6 ms, System: 2.5 ms]
  Range (min … max):     5.8 ms …  11.4 ms    433 runs

Benchmark 2: dist/aqua exec -- jq
  Time (mean ± σ):       6.7 ms ±   0.9 ms    [User: 5.5 ms, System: 2.3 ms]
  Range (min … max):     4.6 ms …  10.4 ms    439 runs

Summary
  dist/aqua exec -- jq ran
    1.18 ± 0.21 times faster than dist/aqua.old exec -- jq
```

If the exe was one that would be found near the start of the list of packages, this doesn't change very much, but it's at least not a regression:

```
$ hyperfine --warmup=20 -N 'dist/aqua'{.old,}' which cmdx'
Benchmark 1: dist/aqua.old which cmdx
  Time (mean ± σ):       5.2 ms ±   0.7 ms    [User: 4.4 ms, System: 1.8 ms]
  Range (min … max):     3.5 ms …   8.4 ms    656 runs

Benchmark 2: dist/aqua which cmdx
  Time (mean ± σ):       5.1 ms ±   0.7 ms    [User: 4.3 ms, System: 1.9 ms]
  Range (min … max):     3.6 ms …   8.4 ms    621 runs

Summary
  dist/aqua which cmdx ran
    1.01 ± 0.19 times faster than dist/aqua.old which cmdx

$ hyperfine --warmup=20 -N 'dist/aqua'{.old,}' exec -- cmdx'
Benchmark 1: dist/aqua.old exec -- cmdx
  Time (mean ± σ):       9.3 ms ±   0.9 ms    [User: 7.9 ms, System: 2.9 ms]
  Range (min … max):     7.5 ms …  15.4 ms    373 runs

Benchmark 2: dist/aqua exec -- cmdx
  Time (mean ± σ):       9.2 ms ±   1.2 ms    [User: 7.5 ms, System: 3.1 ms]
  Range (min … max):     7.2 ms …  20.3 ms    338 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Summary
  dist/aqua exec -- cmdx ran
    1.01 ± 0.17 times faster than dist/aqua.old exec -- cmdx
```

## Check List

- [ ] [Write a GitHub Issue before creating a Pull Request](https://github.com/suzuki-shunsuke/oss-contribution-guide/blob/main/README.md#create-an-issue-before-creating-a-pull-request)
  - Link to the issue:

### DIFF
--- a/pkg/config/registry/package_info_test.go
+++ b/pkg/config/registry/package_info_test.go
@@ -252,7 +252,8 @@ func TestPackageInfo_MaybeHasCommand(t *testing.T) {
 				RepoOwner: "suzuki-shunsuke",
 				RepoName:  "ci-info",
 			},
-		}, {
+		},
+		{
 			title: "potentially empty",
 			has:   []string{"ci-info", "ci-info.prebuilt"},
 			pkgInfo: &registry.PackageInfo{


refi64:
oops, broke aliases

suzuki-shunsuke:
We need to fix CI failures related to [command aliases](https://aquaproj.github.io/docs/guides/command-alias),
but the approach looks good to me.

One of concerns is that when we will add new features like command aliases, this feature may cause a bug if we forget to fix the code.
Maybe it's better to add a method like `ListAvailableCommands` or `HasCommand` to the struct Package or PackageInfo.

refi64:
Redid the implementation and fixed some ugly missing cases, integration tests should pass now. (probably.)

pkg/config/registry/package_info.go
```diff
@@ -586,6 +586,65 @@ func (p *PackageInfo) GetFiles() []*File {
...
+		if doFilesContain(o.Files, exeName, &anyListEmpty) {
+			return true
+		}
+	}
+
+	// If any of the file lists that could be used are empty, then the default
+	// command name would be used, so check that as well.
+	if anyListEmpty && p.defaultCmdName() == exeName {
+		return true
+	}
```
suzuki-shunsuke, line 643: Hmm. I'm wondering this log is really correct.
For instance, in the following case, `p.Files` is empty but the default command name `cli` isn't used.

```yaml
packages:
  - type: github_release
    repo_owner: cli
    repo_name: cli
    version_constraint: "false"
    version_overrides:
      - version_constraint: "true"
        # ...
        files:
          - name: gh
```

https://github.com/aquaproj/aqua/blob/aa48db905ee4c024cd5b14e2b6dd3f9e2778fe25/pkg/config/registry/package_info.go#L664-L680

suzuki-shunsuke: Ah, I see. This function is `Maybe`HasCommand, so this is acceptable.
Even if the package actually doesn't have the command, the package is excluded in subsequent process.

refi64: indeed, in this context false positives are entirely fine

suzuki-shunsuke (approved):

suzuki-shunsuke:
https://github.com/aquaproj/aqua/releases/tag/v2.55.2-0
Looks good.
I'll try this for a while.

suzuki-shunsuke:
🎉 https://github.com/aquaproj/aqua/releases/tag/v2.55.2

suzuki-shunsuke:
There are some ideas to improve the performance.

- https://github.com/aquaproj/aqua/issues/4352
- https://github.com/aquaproj/aqua/issues/4353

PR #4422, opened 2025-12-18
feat: get pseudo-version from Go Proxy if no tagged version exists

### LINKED ISSUE #4421: Add support for installing go packages which have no git tags
### Feature Overview

Some go packages elect not to use git tags for version management, instead relying on go [pseudo-versions](https://go.dev/ref/mod#pseudo-versions) to generate a unique version for each commit.

Aqua should be able to install go packages which only have pseudo-versions.

If we take `golang.org/x/perf/cmd/benchstat` as an example, and add this package to a local registry:

```yaml
packages:
  - type: go_install
    path: golang.org/x/perf/cmd/benchstat
    go_version_path: golang.org/x/perf
```

And attempt to install benchstat with:

```console
$ aqua generate -i -g -s local,golang.org/x/perf/cmd/benchstat
```

We will get a `aqua.yaml` that has no version set:

```yaml
packages:
- name: golang.org/x/perf/cmd/benchstat
  registry: local
  version: "[SET PACKAGE VERSION]"
```

### Example Code

The reason why aqua is unable to find the latest version for benchstat, is that `versiongetter/goproxy/client.go` looks up the `https://proxy.golang.org/$module/@v/list` API endpoint, which does not return pseudo-versions.

If we look at the [go proxy protocol docs](https://go.dev/ref/mod#goproxy-protocol) we see there is another API endpoint that can be used to lookup the latest version of a package including pseudo-versions:

```
https://proxy.golang.org/$module/@latest 	Returns JSON-formatted metadata about the latest known version of a module in the same format as $base/$module/@v/$version.info. The latest version should be the version of the module that the go command should use if $base/$module/@v/list is empty or no listed version is suitable. This endpoint is optional, and module proxies are not required to implement it. 
```

And in fact we can see that the `go` binary when asked to install the latest version of benchstat, will first try the `@v/list` API endpoint followed by `@latest`

```console
$ go install -x golang.org/x/perf/cmd/benchstat@latest 2>&1 | grep proxy.golang.org | grep -E "200|404"
# get https://proxy.golang.org/golang.org/x/perf/@v/list: 200 OK (0.033s)
# get https://proxy.golang.org/golang.org/@v/list: 404 Not Found (0.057s)
# get https://proxy.golang.org/golang.org/x/perf/cmd/@v/list: 404 Not Found (0.060s)
# get https://proxy.golang.org/golang.org/x/perf/cmd/benchstat/@v/list: 404 Not Found (0.062s)
# get https://proxy.golang.org/golang.org/x/perf/@latest: 200 OK (0.030s)
# get https://proxy.golang.org/golang.org/x/@v/list: 404 Not Found (0.064s)
```

### Note

_No response_

gizmoguy:
## Check List

  - Link to the issue:

This PR fixes #4421 by adding the `$module/@v/list` go proxy API endpoint if `@v/list` API response is empty.

After this patch is applied, if you define a local registry like this:

```yaml
packages:
  - type: go_install
    path: golang.org/x/perf/cmd/benchstat
    go_version_path: golang.org/x/perf
```

And attempt to install benchstat with:

```console
$ aqua generate -i -g -s local,golang.org/x/perf/cmd/benchstat
```

We will now get a `aqua.yaml` that has a version set:

```yaml
packages:
- name: golang.org/x/perf/cmd/benchstat@v0.0.0-20251208221838-04cf7a2dca90
  registry: local
```

I also tested and this works for upgrading go packages pinned to a pseudo-version with the `aqua update` command.

### DIFF
--- a/pkg/versiongetter/goproxy/client.go
+++ b/pkg/versiongetter/goproxy/client.go
@@ -26,26 +26,6 @@ func New(client *http.Client) *Client {
 	}
 }
 
-func (c *Client) doHTTPRequest(ctx context.Context, uri string) ([]byte, error) {
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, uri, nil)
-	if err != nil {
-		return nil, fmt.Errorf("create a http request: %w", err)
-	}
-	resp, err := c.client.Do(req)
-	if err != nil {
-		return nil, fmt.Errorf("send a http request: %w", err)
-	}
-	defer resp.Body.Close()
-	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("unexpected status code: %d", resp.StatusCode)
-	}
-	b, err := io.ReadAll(resp.Body)
-	if err != nil {
-		return nil, fmt.Errorf("read a response body: %w", err)
-	}
-	return b, nil
-}
-
 func (c *Client) List(ctx context.Context, logE *logrus.Entry, path string) ([]string, error) {
 	listEndpoint := fmt.Sprintf("https://proxy.golang.org/%s/@v/list", path)
 	b, err := c.doHTTPRequest(ctx, listEndpoint)
@@ -77,4 +57,24 @@ func (c *Client) List(ctx context.Context, logE *logrus.Entry, path string) ([]s
 		return nil, fmt.Errorf("decode the response body as JSON: %w", logerr.WithFields(err, fields))
 	}
 	return []string{payload.Version}, nil
+}
+
+func (c *Client) doHTTPRequest(ctx context.Context, uri string) ([]byte, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, uri, nil)
+	if err != nil {
+		return nil, fmt.Errorf("create a http request: %w", err)
+	}
+	resp, err := c.client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("send a http request: %w", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("unexpected status code: %d", resp.StatusCode)
+	}
+	b, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("read a response body: %w", err)
+	}
+	return b, nil
 }


pkg/versiongetter/goproxy/client.go
```diff
@@ -2,12 +2,19 @@ package goproxy
...
 	"fmt"
 	"io"
 	"net/http"
 	"strings"
+	"time"
 )
 
+type InfoPayload struct {
+	Version string
+	Time    time.Time
```
suzuki-shunsuke:
```suggestion
```

Time is unused.

pkg/versiongetter/goproxy/client.go
```diff
@@ -35,9 +42,31 @@ func (c *Client) List(ctx context.Context, path string) ([]string, error) {
 	if err != nil {
 		return nil, fmt.Errorf("read a response body: %w", err)
 	}
+	return b, nil
+}
+
+func (c *Client) List(ctx context.Context, path string) ([]string, error) {
+	b, err := c.doHTTPRequest(ctx, fmt.Sprintf("https://proxy.golang.org/%s/@v/list", path))
+	if err != nil {
+		return nil, err
```
suzuki-shunsuke: err should be wrapped to distinguish list and latest.

suzuki-shunsuke:
Hmm. It doesn't work.

```console
$ go run ./cmd/aqua g golang.org/x/perf/cmd/benchstat
- name: golang.org/x/perf/cmd/benchstat
  version: '[SET PACKAGE VERSION]'
```

I added `go_version_path: golang.org/x/perf/cmd/benchstat` to https://github.com/aquaproj/aqua-registry/blob/main/pkgs/golang.org/x/perf/cmd/benchstat/registry.yaml , but the result wasn't changed. 🤔

https://aquaproj.github.io/docs/reference/registry-config/go-version-path/

pkg/versiongetter/goproxy/client.go
suzuki-shunsuke: I see. When list API returns 404, doHTTPRequest returns an error, so latest API never calls.
We should continue the process while outputting debug log.

suzuki-shunsuke: Ah, no logger is passed to List method as argument.
Hmm.

suzuki-shunsuke:
```console
$ ~/go/bin/aqua -log-level debug g golang.org/x/perf/cmd/benchstat
DEBU[0000] ignore a package alias in the registry because the alias is duplicate  env=darwin/arm64 package_name= program=aqua program_version=v1.0.0-local registry_name=standard registry_package_alias=&{jreisinger/checkip} registry_package_name=jreisinger/checkip
DEBU[0000] retrieve package versions                     api_endpoint=https://proxy.golang.org/golang.org/x/perf/cmd/benchstat/@v/list env=darwin/arm64 error=unexpected status code: 404 program=aqua program_version=v1.0.0-local repository=/
DEBU[0000] retrieve package versions in 702.626125ms     env=darwin/arm64 program=aqua program_version=v1.0.0-local repository=/
WARN[0000] retrieve package versions                     api_endpoint=https://proxy.golang.org/golang.org/x/perf/cmd/benchstat/@latest env=darwin/arm64 error=list versions: retrieve the latest version: unexpected status code: 404 program=aqua program_version=v1.0.0-local repository=/
- name: golang.org/x/perf/cmd/benchstat
  version: '[SET PACKAGE VERSION]'
```

https://proxy.golang.org/golang.org/x/perf/cmd/benchstat/@latest returns 404.

suzuki-shunsuke:
`/@latest` doesn't return pseudo-version.

```console
$ curl "https://proxy.golang.org/github.com/suzuki-shunsuke/slog-error/@latest"
{"Version":"v0.2.1","Time":"2025-12-05T05:04:05Z","Origin":{"VCS":"git","URL":"https://github.com/suzuki-shunsuke/slog-error","Hash":"c24c9a12709d742f300f1633c079f661812ecf24","Ref":"refs/tags/v0.2.1"}}%                                                                               

$ curl "https://proxy.golang.org/golang.org/x/perf/cmd/benchstat/@latest"      
not found: module golang.org/x/perf/cmd/benchstat: no matching versions for query "latest"% 
```

gizmoguy:
> Hmm. It doesn't work.
> 
> ```
> $ go run ./cmd/aqua g golang.org/x/perf/cmd/benchstat
> - name: golang.org/x/perf/cmd/benchstat
>   version: '[SET PACKAGE VERSION]'
> ```
> 
> I added `go_version_path: golang.org/x/perf/cmd/benchstat` to https://github.com/aquaproj/aqua-registry/blob/main/pkgs/golang.org/x/perf/cmd/benchstat/registry.yaml , but the result wasn't changed. 🤔
> 
> https://aquaproj.github.io/docs/reference/registry-config/go-version-path/

`go_version_path` must be set to the root go module path, so `golang.org/x/perf` for `golang.org/x/perf/cmd/benchstat`

The registry configuration needs to look like:

```
packages:
  - type: go_install
    path: golang.org/x/perf/cmd/benchstat
    go_version_path: golang.org/x/perf
```

pkg/versiongetter/goproxy/client.go
gizmoguy: If the /list API endpoint returns 404, we know the /@latest endpoint will also return 404, so it is not required to query both endpoints.

suzuki-shunsuke:
Ah, I see.

```console
$ curl "https://proxy.golang.org/golang.org/x/perf/@latest"
{"Version":"v0.0.0-20251208221838-04cf7a2dca90","Time":"2025-12-08T22:18:38Z","Origin":{"VCS":"git","URL":"https://go.googlesource.com/perf","Hash":"04cf7a2dca90064b204731fae32fa4e221671c9a"}}%
```

pkg/versiongetter/goproxy/client.go
suzuki-shunsuke: I reverted the change. Thank you. c3473bb524882f62a090ec9662296634f0cb8059

suzuki-shunsuke:
I confirmed it works as expected.

Before:

```console
$ aqua -v
aqua version 2.55.3

$ aqua g golang.org/x/perf/cmd/benchstat
- name: golang.org/x/perf/cmd/benchstat
  version: '[SET PACKAGE VERSION]'
```

After:

```console
$ ~/go/bin/aqua g golang.org/x/perf/cmd/benchstat
- name: golang.org/x/perf/cmd/benchstat@v0.0.0-20251208221838-04cf7a2dca90
```

suzuki-shunsuke:
- https://github.com/aquaproj/aqua-registry/pull/46162

suzuki-shunsuke:
🎉 https://github.com/aquaproj/aqua/releases/tag/v2.56.0

PR #4840, opened 2026-05-20
fix(update-checksum): consider variants when enumerating package assets

### LINKED ISSUE #4837: update-checksum command doesn't consider `variants`
### aqua info

aqua v2.59.0

### Overview

- https://github.com/aquaproj/aqua-registry/issues/54014#issuecomment-4488711629
- https://github.com/aquaproj/aqua/blob/9ede2e816befe0a416dd7a10f6794f8bc9b573b8/pkg/controller/updatechecksum/update.go#L183-L187

### How to reproduce

1. Add packages which use `variants` to aqua.yaml
2. Run `aqua upc`

### Expected behaviour

Checksums for all variants are added to `aqua-checksums.json`.

### Actual behaviour

Only checksums for one variant are added to `aqua-checksums.json`.

### Note

- https://aquaproj.github.io/docs/reference/registry-config/overrides#variants

suzuki-shunsuke:
Close #4837

## Summary

`aqua update-checksum` did not consider `variants` (e.g. `libc: musl` vs `libc: glibc`) when collecting per-runtime checksums. For each of the 6 base GOOS/GOARCH runtimes (`LibC=""`), `pkgInfo.OverrideByRuntime(rt)` returned at most one matching Override, so any sibling Override that differed only by variants was silently dropped — its asset's checksum was never recorded in `aqua-checksums.json`. If every applicable Override for a (GOOS, GOARCH) was variant-constrained (no fallback), the platform's checksum was missing entirely.

This change expands the runtime list per package, based on the Overrides declared by that package, so update-checksum produces a checksum entry for every variant value combination (e.g. one for `libc=musl` and one for `libc=glibc`).

### `pkg/config/registry/override.go`

- Extract `Override.MatchPlatform(rt)` from `Override.Match(rt)`: matches by GOOS, GOArch, and Envs only, ignoring Variants. `Match` now delegates to `MatchPlatform`, then applies the existing variant check. No behavioral change for existing callers.
- Add `SupportedVariantKeys() []string` and `IsSupportedVariantKey(key string) bool` so variant-enumeration code can avoid hard-coding the supported key list (`{"libc"}`).

### `pkg/controller/updatechecksum/variants.go` (new)

- `expandRuntimesByVariants(pkgInfo, rts)`: for each base runtime, finds Overrides that match the (GOOS, GOARCH, Envs) ignoring variants, collects the variant value sets that appear across those Overrides, and emits one Runtime per Cartesian-product combination. Overrides that reference an unsupported variant key are skipped (consistent with `Override.Match` rejecting them). When no matching Override declares variants, the base runtime is returned unchanged — packages without variant-aware Overrides keep their existing behavior.
- Fallback Overrides (no variants) contribute the empty value `""` for each key, so a fallback continues to produce a Runtime with `LibC` cleared alongside the variant-specific Runtimes.

### `pkg/controller/updatechecksum/update.go`

- `updatePackage` now calls `expandRuntimesByVariants(pkg.PackageInfo, rts)` right after `checksum.GetRuntimesFromSupportedEnvs`.
- The `pkgs` map key switches from `rt.Env()` (`"goos/goarch"`) to a new `runtimeKey(rt)` (`"goos/goarch/libc"`), so runtimes differing only by variant no longer collide in the map. `getPkgs`, `allChecksumsCached`, and `prefetchReleaseAssets` are updated to use the new key.
- Per-runtime logging adds a `checksum_libc` field when `LibC` is non-empty for easier debugging; `checksum_env` is preserved.

### Tests

- `pkg/controller/updatechecksum/variants_internal_test.go` (new): covers no-variants no-op, both `libc=musl` and `libc=glibc` expanding for `linux/amd64`, musl-only Override dropping the empty-LibC runtime for that platform, fallback Override coexisting with a libc-constrained Override, Overrides referencing an unsupported variant key being ignored, and a GOOS-only-scoped variant Override spanning matching architectures.
- `pkg/controller/updatechecksum/update_internal_test.go`: updates the existing `TestAllChecksumsCached` to build the `pkgs` map with the new `runtimeKey` instead of `rt.Env()`.

## Notes

- `ChecksumID` is derived from the rendered asset name, so musl and glibc assets naturally produce distinct entries in `aqua-checksums.json`. No changes needed there.
- `install` flows construct `Runtime` from the current host (`runtime.New`), so variants already work correctly for installation; only the multi-runtime enumeration in `update-checksum` needed this fix.

## Test plan

- [x] `cmdx v` (`go vet ./...`)
- [x] `cmdx t` (`go test ./... -race`)
- [x] `cmdx lint` (`golangci-lint run`, 0 issues)
- [ ] Manual: with a `aqua.yaml` referencing a package that declares both `libc: musl` and `libc: glibc` Overrides for `linux/amd64`, confirm that a single `aqua update-checksum` run records checksum entries for both asset names.

## Release Notes

* **New Features**
  * Added support for recognizing supported runtime variant keys.
  * Added variant-aware runtime expansion for checksum updates.

* **Bug Fixes**
  * Improved platform and variant matching.
  * Updated checksum processing and caching to distinguish variants such as libc.
  * Improved fallback handling across runtime combinations.

* **Documentation**
  * Updated the documented aqua CLI version to 2.62.2.

* **Tests**
  * Expanded coverage for variant expansion and checksum caching.


Teko012:
@suzuki-shunsuke Any news on this?

PR #4990, opened 2026-07-09
fix(unarchive): allow benign symlinks that point outside the extraction directory

suzuki-shunsuke:
## Background

The fix for [GHSA-mf5c-hw34-4hpp](https://github.com/aquaproj/aqua/security/advisories/GHSA-mf5c-hw34-4hpp) (commit d5b02b2) rejected any symlink whose target resolves outside the extraction directory and aborted the whole extraction.

This turned out to be too strict and broke the installation of legitimate archives that ship root filesystem images. Concretely, `smol-machines/smolvm` v1.0.3 ships a full Alpine/musl rootfs containing 276 absolute symlinks such as `agent-rootfs/var/run -> /run` and `agent-rootfs/bin/cat -> /bin/busybox`. Installation failed at the very first such symlink:

```
extract files: handling file:
smolvm-1.0.3-darwin-arm64/agent-rootfs/var/run:
the symlink target escapes the extraction directory:
.../smolvm-1.0.3-darwin-arm64/agent-rootfs/var/run -> /run
```

See aquaproj/aqua-registry#55426.

## Why the old check was too strict

Creating a symlink whose target points outside the extraction directory writes nothing outside it — the symlink inode lives inside the directory. The actual threat behind the advisory is a *later* archive entry that follows such a symlink to write outside the directory (e.g. a symlink `pwn -> /outside` followed by a regular file `pwn`, or a file `pwn/x` under a symlinked directory).

I verified against the real smolvm tarball that none of its 276 escaping symlinks are followed by any entry that writes through them, so the archive never writes outside the destination. Rejecting it was a false positive.

## Change

Instead of rejecting the symlink itself, allow it and block the escape at write time:

- `handleSymlink` no longer checks the symlink target; the symlink is created even when its target points outside the extraction directory.
- New `escapesDest` resolves the deepest existing ancestor of a write path (or the path itself) with `filepath.EvalSymlinks` and rejects it when it lands outside the destination. This catches both an escaping symlink planted at the write path and one in a parent directory. The destination is resolved too, so a symlinked temp dir (e.g. macOS `/var -> /private/var`) is not a false positive.
- `escapesDest` is checked before creating parent directories, before `MkdirAll` of a directory entry, and before opening a regular file for writing.
- `HandleFile` is split into `handleDir` / `handleRegularFile` to keep complexity within lint limits.

Net effect: benign escaping symlinks (as in a rootfs image) are preserved and extraction succeeds, while any write that would follow a symlink out of the extraction directory is still rejected.

## Tests

- Keep the existing symlink-traversal and path-traversal tests.
- Add `TestUnarchiver_Unarchive_symlinkDirTraversal`: a regular file whose path descends through a symlinked directory pointing outside is rejected and nothing is written outside.
- Add `TestUnarchiver_Unarchive_symlinkOutsideAllowed`: an archive with a benign escaping symlink (`var/run -> /run`) extracts successfully, the symlink is created, and normal files are still extracted.
- Add a small `buildTarGz` test helper.

Manually verified that the real smolvm-1.0.3 tarball extracts fully with the new code (rootfs symlinks created, `bin/busybox` present), and that the macOS `/var -> /private/var` case does not trigger a false positive.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened archive extraction protections to block path-escape scenarios even when symlinks exist in the destination path.
  * Directories and regular files are now validated at write-time to ensure resolved paths never leave the selected destination.
  * Symlink handling is more permissive at creation time, while still preventing later traversal that could write outside the target.

* **Tests**
  * Improved tar/gzip generation for malicious and valid archives.
  * Added/updated regression coverage for symlink directory traversal, allowed escaping symlinks, and root entry handling.

### DIFF
--- a/pkg/unarchive/archives.go
+++ b/pkg/unarchive/archives.go
@@ -16,10 +16,7 @@ import (
 	"github.com/suzuki-shunsuke/slog-error/slogerr"
 )
 
-var (
-	errEscapeDest        = errors.New("the file path escapes the extraction directory")
-	errSymlinkEscapeDest = errors.New("the symlink target escapes the extraction directory")
-)
+var errEscapeDest = errors.New("the file path escapes the extraction directory")
 
 type handler struct {
 	fs       afero.Fs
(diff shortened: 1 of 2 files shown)


KSHITIZ6341:
Post-merge follow-up on `pkg/unarchive/archives.go:138`: I may be missing a later follow-up, but this looks worth checking. The write-time guard seems to miss the case where the escaping symlink's final target does not exist yet.

In a local repro against this PR branch, an archive with `pwn -> <outsideDir>/created-outside-target` followed by a regular file entry named `pwn` returned success and created `<outsideDir>/created-outside-target` with the archive payload. Since `filepath.EvalSymlinks(p)` fails when the target leaf does not exist yet, `escapesDest` walks back to the parent under `dest`; the later `OpenFile(..., O_CREATE|O_WRONLY)` then follows the symlink and creates the outside file.

Suggested direction: resolve/check existing symlink components without requiring the final target to exist, and add a regression test for the missing-outside-target variant.

suzuki-shunsuke:
@KSHITIZ6341 Thank you for the report.
You're right.
I've created a pull request to resolve the issue.
- https://github.com/aquaproj/aqua/pull/4992

PR #4992, opened 2026-07-09
fix(unarchive): block writes through a dangling escaping symlink

suzuki-shunsuke:
## Summary

Follow-up to #4990. KSHITIZ6341 reported (https://github.com/aquaproj/aqua/pull/4990#issuecomment-4927707857) that the write-time escape guard still lets an archive create a file outside the extraction directory when the escaping symlink's target does not exist yet.

I reproduced it against `main`: an archive with

```
pwn -> <outsideDir>/created-outside-target   (symlink, target does NOT exist)
pwn                                           (regular file)
```

extracts successfully and creates `<outsideDir>/created-outside-target` with the archive payload — a write outside `$(aqua root-dir)`.

## Root cause

`escapesDest` resolved paths with `filepath.EvalSymlinks`, which returns an error when a symlink's final target does not exist. For a dangling symlink planted at `dstPath`, `EvalSymlinks(dstPath)` failed, so the guard walked back to the parent directory (inside `dest`), concluded there was no escape, and let the write proceed. `OpenFile(dstPath, O_CREATE|O_WRONLY)` then followed the dangling symlink and created the target file outside `dest`.

The already-covered variants stayed safe by accident: when the target exists, `EvalSymlinks` resolves it and the escape is caught; and a dangling symlink used as a parent directory (`pwn/file`) makes `MkdirAll` fail so the entry is skipped. Only a dangling symlink at the write leaf slipped through.

## Fix

Replace the `EvalSymlinks`-only walk in `escapesDest` with `resolveSymlinks`, which follows symlink components manually via `os.Lstat` / `os.Readlink` and does not require the final target to exist. A dangling symlink's target is resolved (relative targets against the link's directory) and checked against `dest`, so an escaping target is detected whether or not it already exists. A symlink-hop counter (`maxSymlinkHops = 255`) guards against cycles.

Behavior for the existing cases is unchanged:
- Benign escaping symlinks (rootfs images like `var/run -> /run`) are still created and extraction still succeeds — symlink entries never go through `escapesDest`.
- Path traversal (`..`), the `./` root entry, and traversal through an existing escaping symlink are all still handled as before.

## Affected versions

No released version is affected by this bug; it exists only on `main`.

- This is a regression introduced by #4990, which is not part of any release tag yet (`git tag --contains` of the merge commit returns nothing; it is not an ancestor of the latest tag `v2.60.2-1`).
- Before #4990, `handleSymlink` rejected any symlink whose target resolves outside `dest` at creation time (`symlinkTargetWithinDest`), aborting extraction. A dangling escaping symlink like `pwn -> <outside>/missing` was therefore blocked regardless of whether its target existed, so this attack path did not exist.
- The original GHSA-mf5c-hw34-4hpp fix (d5b02b2) ships in `v2.60.1` and later, which retain that strict "reject the escaping symlink at creation" behavior and are not vulnerable to this variant.

Because #4990 is unreleased, #4990 and this fix will land in the same release, so end users are never exposed. No emergency release or advisory update is required.

## Tests

Add `TestUnarchiver_Unarchive_danglingSymlinkTraversal` covering the missing-target variant. It fails before this change (the outside file is created) and passes after. All existing unarchive tests continue to pass, and `golangci-lint` is clean.

Reported by @KSHITIZ6341.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved protection against archive extraction paths escaping the destination directory.
  * Better handling of symlinks, including dangling links, parent-directory links, and symlink loops.
  * Extraction now fails safely when a path would resolve outside the intended location, helping prevent unwanted file creation.

### DIFF
--- a/pkg/unarchive/archives.go
+++ b/pkg/unarchive/archives.go
@@ -122,34 +122,70 @@ func (h *handler) handleSymlink(dstPath, target string) {
 	}
 }
 
-// escapesDest reports whether path resolves outside h.dest once the symlinks in
-// its already-existing portion are followed. It resolves the deepest existing
-// ancestor of path (or path itself), so it detects both a symlink planted at
-// path and an escaping symlink in any of its parent directories. h.dest is
-// resolved too because it may itself contain symlinks (e.g. macOS /var ->
-// /private/var), which would otherwise make every entry look like an escape.
+// maxSymlinkHops bounds how many symlinks escapesDest follows before giving up,
+// guarding against symlink cycles. It matches the conventional MAXSYMLINKS.
+const maxSymlinkHops = 255
+
+// escapesDest reports whether writing to path would land outside h.dest once the
+// symlinks along path are followed. Unlike filepath.EvalSymlinks it does not
+// require path's final target to exist, so it also catches a dangling symlink
+// planted at path that a later O_CREATE write would follow out of h.dest. It
+// detects a symlink planted at path itself as well as an escaping symlink in any
+// parent directory. h.dest is resolved too because it may itself contain
+// symlinks (e.g. macOS /var -> /private/var), which would otherwise make every
+// entry look like an escape.
 func (h *handler) escapesDest(path string) bool {
 	dest, err := filepath.EvalSymlinks(h.dest)
 	if err != nil {
 		dest = filepath.Clean(h.dest)
 	}
-	p := path
-	for {
-		resolved, err := filepath.EvalSymlinks(p)
-		if err == nil {
-			resolved = filepath.Clean(resolved)
-			if resolved == dest {
-				return false
-			}
-			return !strings.HasPrefix(resolved, dest+string(filepath.Separator))
-		}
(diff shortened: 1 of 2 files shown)


suzuki-shunsuke:
@coderabbitai Thanks for the suggestion. We'll keep these three as separate tests rather than consolidating them into a table-driven one.

The commonality is limited to building the archive and asserting that Unarchive returns an error, which is already shared via the `buildTarGz` helper. The parts that differ span both setup and assertion:

- `symlinkTraversal`: target is an existing file; asserts the file's content is left unchanged.
- `symlinkDirTraversal`: target is an existing directory; asserts no file is created under it.
- `danglingSymlinkTraversal`: target does not exist yet; asserts the path is not created.

A table would therefore need per-row setup and assertion closures rather than simple input/expected values, which does not read more clearly than the current form. For security regression tests we prefer each attack scenario to be self-contained and readable on its own, so we're leaving them as is.

PR #5002, opened 2026-07-11
feat(unarchive): fall back to the system tar for GNU sparse tarballs

### LINKED ISSUE #5001: Support extracting GNU sparse (PAX GNU.sparse.*) tarballs (e.g. smol-machines/smolvm)
## Overview

aqua fails to extract release tarballs that contain GNU sparse (PAX `GNU.sparse.*`) files. We should decide how to handle this — the most likely fix is to detect sparse-PAX tar archives and fall back to the system `tar`, as mise did.

Concrete failing package: `smol-machines/smolvm` (attempted in aqua-registry). Installing it fails with:

```
WRN copy a file ... error="archive/tar: sparse file contains unreferenced data"
ERR install the package ... error="extract files: archive/tar: sparse file contains unreferenced data"
```

Reproduces on both `smolvm-1.1.1` and the latest `smolvm-1.3.8` (`*-darwin-arm64.tar.gz`).

## Root cause

aqua extracts archives via `mholt/archives`, which uses the Go standard library `archive/tar`. The smolvm tarball stores large disk-image templates (e.g. `storage-template.ext4`, logical 20 GiB) as GNU sparse format 1.0. Its PAX extended header looks like this:

```
20 size=21474836480
22 GNU.sparse.major=1
22 GNU.sparse.minor=0
67 GNU.sparse.name=smolvm-1.3.8-darwin-arm64/storage-template.ext4
35 GNU.sparse.realsize=21474836480
```

The problem is the `size` record. For GNU sparse 1.0, the PAX `size` keyword must be the physical size stored in the archive (the sparse map + data blocks — about 24.5 MiB here; the ustar header's own size field is correctly `25723392`). This archive instead sets `size` to the logical/real size (`21474836480`). Go's `archive/tar` trusts the PAX `size` as the physical byte count, so after reading the real data it still believes ~20 GiB of physical data remain and returns `errUnrefData` ("sparse file contains unreferenced data").

I verified the cause directly:
- Real `GNU tar 1.35` writing a 20 GiB sparse file in posix (1.0) format emits **no** `size=` record (only `GNU.sparse.realsize`); Go reads it fine.
- `bsdtar`/libarchive also omits `size=`; Go reads it fine.
- Injecting only a `size=<realsize>` record into an otherwise-valid GNU-tar sparse archive reproduces the exact same `archive/tar: sparse file contains unreferenced data`. GNU tar and bsdtar still extract that injected archive without complaint.

So the archive is non-standard (no correct tar produces this `size` record), but GNU tar / bsdtar tolerate it while the strict Go and Rust extractors do not.

## Upstream will not fix it

The same issue was reported against mise: smol-machines/smolvm#506. The smolvm maintainer closed it as `wontfix`, considering the sparse files intentional and the extractor at fault. So we cannot expect the release tarball to be repacked.

## How mise solved it

mise hit the identical wall — its Rust `tar` crate (0.4.46) also fails on these PAX `GNU.sparse.*` archives. Fix: jdx/mise#10821 detects PAX `GNU.sparse.*` metadata before extracting and falls back to the system `tar` for sparse archives (mise already had a system-`tar` fallback path). Requires `tar` on `PATH` (standard on macOS/Linux).

## Options for aqua

1. Add a system-`tar` fallback for GNU sparse PAX archives in `pkg/unarchive` (mise's approach). Pros: makes such packages installable, and system `tar` preserves on-disk sparseness. Cons: aqua currently avoids depending on external `tar`; needs a `tar` on `PATH` and a Windows story.
2. Do nothing in aqua; don't add such packages to aqua-registry (skip/exclude the affected assets/platforms).
3. Something else (e.g. a custom tar reader that ignores the bogus PAX `size`).

Additional consideration: even if `archive/tar` could read it, aqua's current extraction does `io.Copy` into a regular file, which would materialize the holes as zeros — ~20 GiB + ~10 GiB written to disk for smolvm. A system-`tar` fallback would also avoid that, which is another point in favor of option 1.

## Repro

```console
$ curl -sL -O https://github.com/smol-machines/smolvm/releases/download/v1.3.8/smolvm-1.3.8-darwin-arm64.tar.gz
# minimal Go repro
$ cat > read.go <<'EOF'
package main
import ("archive/tar";"compress/gzip";"fmt";"io";"os")
func main(){
	f,_:=os.Open(os.Args[1]); gz,_:=gzip.NewReader(f); tr:=tar.NewReader(gz)
	for{ h,e:=tr.Next(); if e==io.EOF{return}; if e!=nil{fmt.Println("Next:",e);return}
		_,ce:=io.Copy(io.Discard,tr); fmt.Printf("%s err=%v\n",h.Name,ce); if ce!=nil{return} }
}
EOF
$ go run read.go smolvm-1.3.8-darwin-arm64.tar.gz
smolvm-1.3.8-darwin-arm64/ err=<nil>
smolvm-1.3.8-darwin-arm64/storage-template.ext4 err=archive/tar: sparse file contains unreferenced data
```

suzuki-shunsuke:
## Overview

aqua fails to extract release tarballs that store files in GNU sparse format (PAX `GNU.sparse.*`). A concrete case is `smol-machines/smolvm`, whose darwin release archives contain multi-GiB sparse ext4 disk templates. Installation fails with:

```
ERR install the package ... error="extract files: archive/tar: sparse file contains unreferenced data"
```

This makes such packages uninstallable via aqua. Refs #5001.

## Root cause

aqua extracts archives via `mholt/archives`, which uses the Go standard library `archive/tar`. Go's `archive/tar` cannot reliably extract GNU sparse archives — GNU tar and bsdtar can. In smolvm's case the tarball is a nonstandard GNU sparse 1.0 archive whose PAX header carries `size=<realsize>` (the logical size) instead of the physical stored size, and Go trusts PAX `size` as the physical byte count and errors after reading the logical data. Go's Rust counterpart (`mise`) hit the identical wall and solved it the same way (`jdx/mise#10821`). Upstream closed the equivalent report as `wontfix` (`smol-machines/smolvm#506`), so the archive will not be repacked.

## Change

Detect a GNU sparse entry during the normal extraction walk and fall back to the system `tar`:

- In `handler.HandleFile`, detect a sparse entry (a `GNU.sparse.*` PAX record, or the old GNU sparse `Typeflag`) **before its body is read** and abort the walk with `fs.SkipAll`. Detecting before the body read is important: reading a sparse body would materialize its full logical size as zeros (tens of GiB for smolvm) before failing.
- `handler.unarchive` then re-extracts the whole archive with `tar -x -f <file> -C <dest>` through the existing `Executor` (the same mechanism the `dmg`/`pkg` unarchivers already use). The system `tar` auto-detects the compression and restores the files sparsely on disk instead of writing the holes as zeros.
- The `handler` now receives the `Executor` from `getUnarchiver` (previously only `dmg`/`pkg` did).

Detection is structural (any GNU sparse entry), so well-formed sparse archives are also routed through the system `tar`, which additionally preserves on-disk sparseness.

### Notes / trade-offs

- The system `tar` path does not apply aqua's custom symlink-escape / path-traversal guards from `archives.go`; it is scoped to the narrow GNU sparse case and relies on `tar`'s own default protections (both GNU tar and bsdtar reject `..`/absolute members and do not follow symlinks on extract; no `-P` is passed). This matches mise's design.
- A `tar` binary on `PATH` is required for GNU sparse archives (standard on macOS and Linux); otherwise a clear error is returned.
- The fallback relies on the system `tar` being able to read the archive. GNU tar cannot fully extract smolvm's specific malformation — only bsdtar can — but smolvm's asset is a macOS (darwin) asset, where the system `tar` is bsdtar, so it works there.

## Tests

- Adds `pkg/unarchive/testdata/gnu-sparse.tar.gz`, a small well-formed GNU sparse 1.0 fixture (member `sparse.img`, logical size 131072, `HELLO` at offset 0, hole after) that GNU tar, bsdtar, and Go's reader all handle.
- Adds `TestUnarchiver_Unarchive_gnuSparse`, which extracts the fixture through the public `Unarchiver` with a real executor and asserts the sparse file is extracted with the correct size and content. It is skipped when no `tar` is on `PATH`.

## Verification

- `cmdx v` (`go vet ./...`), `cmdx t` (`go test ./... -race`), and `cmdx l` (`golangci-lint run`) all pass.
- End-to-end against the real `smolvm-1.3.8-darwin-arm64.tar.gz`: the fallback warning is logged and all 705 entries extract (130 dirs, 368 files, 337 symlinks); `storage-template.ext4` is written at its 21 GiB logical size and stays sparse on disk.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved extraction of GNU sparse tar archives.
  * Preserves sparse file sizes, content, and zero-filled holes during extraction.
  * Automatically uses a compatible system extraction method when needed.


suzuki-shunsuke:
- https://github.com/aquaproj/aqua-registry/pull/56837/changes/1daf421162ab9a36d399a5d3bfeb9445f3e1dd86
- https://github.com/aquaproj/aqua/releases/tag/v2.61.0-0

PR #5048, opened 2026-07-21
fix(unarchive): expand a pkg file into a non-existent destination

### LINKED ISSUE #5040: pkg-format packages fail to install on macOS with "File exists" (regression in v2.62.0 from #5025)
## aqua info

```console
$ aqua info
{
  "version": "2.62.0",
  "os": "darwin",
  "arch": "arm64",
  "cwd": "/Users/(USER)/repo",
  "root_dir": "/Users/(USER)/.local/share/aquaproj-aqua",
  "env": {
    "AQUA_GENERATE_WITH_DETAIL": "true",
    "AQUA_GLOBAL_CONFIG": ":/Users/(USER)/.config/aquaproj-aqua/aqua.yaml"
  },
  "config_files": [
    {
      "path": "/Users/(USER)/repo/aqua.yaml"
    }
  ]
}
```

This issue occurs with aqua v2.62.0 but doesn't occur with v2.61.0.

## Overview

Since v2.62.0, installing any package with `format: pkg` on macOS fails with `Could not unarchive ... (File exists)`. For example `aws/aws-cli`, which uses the `.pkg` asset on darwin.

Cause: #5025 changed `Installer.unarchive` to extract into a temporary directory and rename it into place:

```go
tempDir, err := afero.TempDir(is.fs, tempDir, "") // creates the directory
...
is.unarchiver.Unarchive(ctx, logger, &unarchive.File{...}, tempDir)
```

`afero.TempDir` **creates** the directory, and `pkgUnarchiver` passes it straight to pkgutil:

```go
u.executor.ExecAndOutputWhenFailure(osexec.Command(ctx, "pkgutil", "--expand-full", tempFilePath, u.dest))
```

`pkgutil --expand-full` requires the destination to **not exist**, and fails with `File exists` otherwise (its error message names the source file, which makes the log misleading). Reproducible outside aqua:

```console
$ pkgutil --expand-full AWSCLIV2-2.36.2.pkg fresh-dir      # ok
$ mkdir existing-dir
$ pkgutil --expand-full AWSCLIV2-2.36.2.pkg existing-dir
Could not unarchive AWSCLIV2-2.36.2.pkg (The operation couldn’t be completed. File exists)
```

Before #5025 the destination passed to the unarchiver was the final install path, which never existed at that point, so pkgutil succeeded.

## How to reproduce

aqua.yaml

```yaml
registries:
  - type: standard
    ref: v4.539.0
packages:
  - name: aws/aws-cli@2.36.2
```

Executed command and output

```console
$ aqua i
INF download and unarchive the package program=aqua version=2.62.0 env=darwin/arm64 package_name=aws/aws-cli package_version=2.36.2 registry=standard
Could not unarchive /var/folders/5r/.../T/624730322 (The operation couldn’t be completed. File exists)

ERR install the package program=aqua version=2.62.0 env=darwin/arm64 package_name=aws/aws-cli package_version=2.36.2 registry=standard error="unarchive a pkg format file: exit status 1"
ERR aqua failed program=aqua version=2.62.0 env=darwin/arm64 config_file_path=/Users/(USER)/repo/aqua.yaml error="install packages: it failed to install some packages"
```

(The package must not already be present under `$AQUA_ROOT_DIR/pkgs`.)

## Expected behaviour

The package installs successfully, as with v2.61.0.

## Actual behaviour

Installation fails with `unarchive a pkg format file: exit status 1` / `Could not unarchive ... (File exists)`.

## Note

Possible fix: since pkgutil needs a non-existent destination, `pkgUnarchiver` could remove the (empty) destination directory before invoking pkgutil, or expand into a child path such as `filepath.Join(dest, "pkg")`.

Workaround for affected users: pin aqua to v2.61.0, or manually expand the pkg into the install path (aqua treats its existence as installed):

```console
$ curl -LO https://awscli.amazonaws.com/AWSCLIV2-2.36.2.pkg
$ shasum -a 256 AWSCLIV2-2.36.2.pkg   # verify against aqua-checksums.json
$ pkgutil --expand-full AWSCLIV2-2.36.2.pkg \
    "$HOME/.local/share/aquaproj-aqua/pkgs/http/awscli.amazonaws.com/AWSCLIV2-2.36.2.pkg"
$ aqua i
```

suzuki-shunsuke:
## Overview

Fixes #5040.

Since v2.62.0, installing any package with `format: pkg` on macOS fails with `Could not unarchive ... (File exists)` / `unarchive a pkg format file: exit status 1`.

#5025 changed `Installer.unarchive` to extract into a temporary directory and rename it into place. `afero.TempDir` creates that directory, and `pkgUnarchiver` passes it straight to `pkgutil --expand-full`, which requires the destination to not exist. Before #5025 the destination was the final install path, which never existed at that point, so pkgutil created it itself.

## How to fix

`pkgUnarchiver` removes the destination before running pkgutil. `Remove` only succeeds on an empty directory, so a destination that holds files is reported as an error rather than being discarded.

Expanding into a child path such as `filepath.Join(dest, "pkg")` was rejected: the destination is the package root that registry `files[].src` values are resolved against (e.g. `bep/s3deploy` declares `src: Payload/s3deploy`), so nesting one level deeper would break every pkg format package.

`dmg` is unaffected: `dmgUnarchiver` creates the destination itself and `osfile.Copy` merges into an existing directory.

## Tests

Added `TestUnarchiver_Unarchive_pkg`, which asserts the destination does not exist when the command is invoked, and `TestUnarchiver_Unarchive_pkg_destNotEmpty`, which asserts a non-empty destination is not removed. They use a stub `Executor`, so they run on Linux CI too. Both fail without this change.

Verified end to end on macOS (darwin/arm64) with a throwaway root directory:

- `bep/s3deploy@v2.16.0` (pkg): fails with the reported error on `main`, installs successfully with this change, and lands at `pkgs/.../s3deploy_2.16.0_darwin-universal.pkg/Payload/s3deploy` as before.
- `99designs/aws-vault@v7.2.0` (dmg): still installs and runs.
- The `temp` directory is left empty in both cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved package archive extraction by ensuring the destination is clear before expansion.
  * Prevented existing destination contents from being overwritten when extraction cannot safely proceed.
  * Added clearer error handling for failures while preparing the destination directory.

* **Tests**
  * Added coverage for extraction into new destinations and protection of non-empty destinations.

### DIFF
--- a/pkg/unarchive/pkg.go
+++ b/pkg/unarchive/pkg.go
@@ -2,7 +2,9 @@ package unarchive
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io/fs"
 	"log/slog"
 	"path/filepath"
 
@@ -27,6 +29,14 @@ func (u *pkgUnarchiver) Unarchive(ctx context.Context, _ *slog.Logger, src *File
 	tempFilePath, err := src.Body.Path()
 	if err != nil {
 		return fmt.Errorf("get a temporary file path: %w", err)
+	}
+
+	// pkgutil --expand-full fails with "File exists" unless the destination is
+	// absent, but the caller hands over a directory it has already created.
+	// Remove only succeeds on an empty directory, so a populated destination is
+	// never destroyed here.
+	if err := u.fs.Remove(u.dest); err != nil && !errors.Is(err, fs.ErrNotExist) {
+		return fmt.Errorf("remove the destination directory before expanding a pkg file: %w", err)
 	}
 
 	if _, err := u.executor.ExecAndOutputWhenFailure(osexec.Command(ctx, "pkgutil", "--expand-full", tempFilePath, u.dest)); err != nil {
--- a/pkg/unarchive/unarchive_test.go
+++ b/pkg/unarchive/unarchive_test.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"fmt"
 	"io"
 	"log/slog"
 	"net/http"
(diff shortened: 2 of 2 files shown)


suzuki-shunsuke:
## Before

```console
$ container --help
Jul 21 23:05:05.306 INF download and unarchive the package program=aqua version=v1.0.0-local env=darwin/arm64 exe_name=container package_name=apple/container package_version=1.1.0 registry=standard
Downloading apple/container 1.1.0 100% |██████████████████████████████████████████████████████████| (90/90 MB, 9.7 MB/s)
Could not unarchive /var/folders/fc/1bgyy3_d3x90m_t04qbw5f8m0000gn/T/082901251 (The operation couldn’t be completed. File exists)

Jul 21 23:05:16.147 ERR aqua failed program=aqua version=v1.0.0-local env=darwin/arm64 exe_name=container package_name=apple/container package_version=1.1.0 registry=standard error="install the package: unarchive a pkg format file: exit status 1"
```

PR #5073, opened 2026-07-26
fix: update the tfcmt registry pin in the cosign test

suzuki-shunsuke:
Fixes the CI failure on #5046.

## Problem

`integration-test-all-envs` fails while installing `suzuki-shunsuke/tfcmt@v4.14.17`:

```
Error: loading verifier from key opts: loading cert: loading URL
https://github.com/suzuki-shunsuke/tfcmt/releases/download/v4.14.17/tfcmt_4.14.17_checksums.txt.pem:
server returned HTTP 404
```

## Cause

Releases produced by `suzuki-shunsuke/go-release-workflow` now publish a single Sigstore bundle (`*_checksums.txt.sigstore.json`) instead of the separate `.pem`, `.sig` and `.bundle` assets. `tests/cosign/aqua.yaml` pinned the `tfcmt` registry at `501377ba`, which predates the registry entry that verifies with the bundle, so cosign kept asking for an asset that no longer exists.

## Change

Pins the `tfcmt` registry at `52c9bfe6` (aqua-registry v4.543.0), whose `tfcmt` entry uses `cosign.bundle` with `tfcmt_{{trimV .Version}}_checksums.txt.sigstore.json`. The `TODO update` marker is kept, and the tag it corresponds to is now recorded in the comment.

The other two pins are left alone, since they verify packages that are unaffected.

This targets the Renovate branch rather than `main`, so review it here and it will land with #5046.

### DIFF
--- a/tests/cosign/aqua.yaml
+++ b/tests/cosign/aqua.yaml
@@ -13,7 +13,7 @@ registries:
     ref: 98524ca5b420af9115275d110b2cadeab60f49e2 # TODO update
     name: soulshack
   - type: standard
-    ref: 52c9bfe694d3a246691ad17fa5718b74c806a56e # v4.543.0 TODO update
+    ref: 52c9bfe694d3a246691ad17fa5718b74c806a56e # v4.543.0
     name: tfcmt
 packages:
   - name: suzuki-shunsuke/mkghtag@v0.1.11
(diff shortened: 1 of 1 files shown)


tests/cosign/aqua.yaml
```diff
@@ -13,7 +13,7 @@ registries:
     ref: 98524ca5b420af9115275d110b2cadeab60f49e2 # TODO update
     name: soulshack
   - type: standard
-    ref: 501377ba89acc447dce6b7ad69f51f705d00bff0 # TODO update
+    ref: 52c9bfe694d3a246691ad17fa5718b74c806a56e # v4.543.0 TODO update
```
suzuki-shunsuke:
```suggestion
    ref: 52c9bfe694d3a246691ad17fa5718b74c806a56e # v4.543.0
```

PR #5114, opened 2026-08-13
docs: fix the broken star history chart

Dessalines39394:
The star history chart in the README is currently broken and does not render. This change points the chart at a working server with the same data source, so the chart is displayed correctly again.

The badge is left unchanged.

## Summary by CodeRabbit

* **Documentation**
  * Updated the Star History Chart badge and link to use the current Star History website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Declarative CLI Version Manager written in Go.
 - Secure
 - Easy to use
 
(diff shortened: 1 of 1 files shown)


suzuki-shunsuke:
I wanted to merge this PR, but the commit wasn't signed so we can't accept this.

PR #5130, opened 2026-08-19
fix: filter prerelease tags in github_tag version getter

AruneshDwivedi:
filterTag in github_tag.go did not auto-reject prerelease tags the way filterRelease does in github_release.go. When more than PerPage prerelease tags accumulate before a stable release, Get() short-circuits on page 1 and returns the highest prerelease instead of finding the stable version on page 2.

Adds a prerelease check to filterTag and a regression test for the pagination edge case.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prerelease tags are now excluded when selecting stable versions, even when they appear before the stable release in tag listings.
  * Version lookup now handles paginated tag results correctly.
  * Requests beyond the available tag pages return an empty result instead of failing.

* **Tests**
  * Added coverage for stable version selection after multiple prerelease tags.

### DIFF
--- a/pkg/versiongetter/github_tag.go
+++ b/pkg/versiongetter/github_tag.go
@@ -105,6 +105,10 @@ func (g *GitHubTagVersionGetter) List(ctx context.Context, logger *slog.Logger, 
 
 func filterTag(logger *slog.Logger, tag *github.RepositoryTag, filters []*Filter) bool {
 	tagName := tag.GetName()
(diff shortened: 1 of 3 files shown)


AruneshDwivedi (commented):
.

<sub>Commented in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/aquaproj/aqua/pull/5130)</sub>

PR #5162, opened 2026-08-29
test: pin the pnpm test registry to the signer_workflow fix

suzuki-shunsuke:
Fixes the `integration-test-all-envs` failure on #5096.

## Why it fails

This PR's branch bumps the embedded `cli/cli` from v2.96.0 to v2.97.0. v2.97.0 fixed a security issue by applying `regexp.QuoteMeta` to `--signer-workflow`, so the value is now matched literally instead of as a regex:

```go
// v2.96.0 — pkg/cmd/attestation/verify/policy.go
return fmt.Sprintf("^https://%s/%s", hostname, signerWorkflow), nil
// v2.97.0
return "^" + regexp.QuoteMeta(fmt.Sprintf("https://%s/%s", hostname, signerWorkflow)), nil
```

Released aqua-registry versions store the value pre-escaped for the old behaviour:

```yaml
signer_workflow: pnpm/pnpm/\.github/workflows/release\.yml
```

`QuoteMeta` escapes that `\.` again, producing a matcher that requires a literal backslash in the certificate SAN. It can never match, so pnpm fails verification:

```
Error: verifying with issuer "sigstore.dev"
ERR install the package  package_name=pnpm/pnpm package_version=v11.5.2
    error="verify the asset: verify a package with gh attestation: verify GitHub Artifact Attestations"
```

## This change

aquaproj/aqua-registry#59621 removed the escaping from all 38 affected packages, but it is not in a release yet — v4.557.0 was cut just before it landed.

So `tests/pnpm/aqua.yaml` pins the registry `ref` to that commit (`19070e42149f21efadd310d28c20caa9496fb105`) until a release contains it, and the SHA is added to the standard-registry allowlist in `aqua/aqua-policy.yaml`. `tests/cosign/aqua.yaml` already pins SHAs the same way, with the same policy allowlist, so this follows the existing pattern rather than introducing one.

`tests/pnpm/aqua-checksums.json` is regenerated with `aqua update-checksum`. Besides the new registry checksum it picks up the `linux-arm64-musl` and `linux-x64-musl` entries the newer registry added for pnpm.

The `# renovate:` annotation is replaced with a TODO to restore it, since Renovate cannot track a SHA.

## Verification

Ran the test locally on darwin/arm64 against this branch's build, so with the v2.97.0 gh this PR pins:

```
INF verify GitHub Artifact Attestations  package_name=pnpm/pnpm package_version=v11.5.2
INF create a symbolic link  file_name=pn
INF creating a hard link  file_name=pnpx
```

Verification passes and the install completes.

### DIFF
--- a/aqua/aqua-policy.yaml
+++ b/aqua/aqua-policy.yaml
@@ -3,7 +3,7 @@
 # aqua Policy
 registries:
   - type: standard
-    ref: semver(">= 3.0.0") or Version in ["f9cce37273a70e2f5664fb4c3708169ffe7e320c", "1f20d0c2211df45694e762fcb20830d5c3cedf95", "98524ca5b420af9115275d110b2cadeab60f49e2", "501377ba89acc447dce6b7ad69f51f705d00bff0"]
+    ref: semver(">= 3.0.0") or Version in ["f9cce37273a70e2f5664fb4c3708169ffe7e320c", "1f20d0c2211df45694e762fcb20830d5c3cedf95", "98524ca5b420af9115275d110b2cadeab60f49e2", "501377ba89acc447dce6b7ad69f51f705d00bff0", "19070e42149f21efadd310d28c20caa9496fb105"]
   - name: main
     type: local
     path: ../tests/main/registry.yaml
--- a/tests/pnpm/aqua-checksums.json
+++ b/tests/pnpm/aqua-checksums.json
@@ -6,8 +6,18 @@
       "algorithm": "sha256"
     },
     {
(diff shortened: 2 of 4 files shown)


suzuki-shunsuke:
Superseded by #5163, closing.

#5163 makes aqua undo the regex escaping before handing the value to gh, so registry entries written for the old behaviour keep verifying. That fixes the same test failure without pinning the registry to a SHA, and it also stops #5096 from being a breaking change for users on a registry version released before aquaproj/aqua-registry#59621 — which was the larger problem this PR did not address.

Verified: with #5163, `tests/pnpm` passes on the existing `ref: v4.546.0`, escaped value and all.

PR #5168, opened 2026-08-29
fix: update the pnpm test checksums for v11.24.0

suzuki-shunsuke:
Fixes the `integration-test-all-envs (ubuntu-24.04-arm)` failure on #5071.

`tests/pnpm/aqua.yaml` moves to pnpm v11.24.0, but `tests/pnpm/aqua-checksums.json` still listed v11.5.2. With `require_checksum: true` aqua refuses to install a package it has no checksum for:

```
ERR install the package  package_name=pnpm/pnpm package_version=v11.24.0 registry=standard
    doc=https://aquaproj.github.io/docs/reference/codes/001
    error="checksum is required"
```

Regenerated with `aqua update-checksum --prune`. Alongside the version bump it picks up two assets the registry added since:

```diff
+  pnpm-linux-arm64-musl.tar.gz
+  pnpm-linux-x64-musl.tar.gz
```

which is why arm64 was the environment that failed.

The registry checksum entry also moves from v4.523.0 to v4.547.0, matching the `ref` already set in `aqua.yaml`.

## Verification

Ran the test locally on darwin/arm64 against this branch:

```
INF verify GitHub Artifact Attestations  package_name=pnpm/pnpm package_version=v11.24.0
INF create a symbolic link  file_name=pn
INF creating a hard link  file_name=pnpx
```

Installation completes with checksum verification enabled.

Base is the Renovate branch, so merging this puts the fix on #5071.

### DIFF
--- a/tests/pnpm/aqua-checksums.json
+++ b/tests/pnpm/aqua-checksums.json
@@ -1,33 +1,43 @@
 {
   "checksums": [
     {
-      "id": "github_release/github.com/pnpm/pnpm/v11.5.2/pnpm-darwin-arm64.tar.gz",
-      "checksum": "54993DAE26BEA0F3C1B0E15F9427F6F6A86827D56F32D1D1554D8CDA59A62399",
+      "id": "github_release/github.com/pnpm/pnpm/v11.24.0/pnpm-darwin-arm64.tar.gz",
+      "checksum": "C4DF428B6C348B3037BA851359215065CE070D711DC96A9750052FDBFBE240D3",
       "algorithm": "sha256"
     },
     {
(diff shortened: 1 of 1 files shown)


suzuki-shunsuke:
No longer needed — closing.

#5169 added `tests/pnpm` to the autofix job, and after #5071 was updated with main, autofix.ci regenerated the checksums itself in `328bf36b` (`tests/pnpm/aqua-checksums.json +14/-14`). That is exactly the change this PR was making by hand.

Which is the better outcome: rather than a one-off fix, the next pnpm bump will not need one at all.

PR #4871, opened 2026-05-27
chore: try mholt/archives v0.1.6-0.20260526201101-89571374212f

suzuki-shunsuke:
- https://github.com/aquaproj/aqua/issues/4639#issuecomment-4554482143

PR #5171, opened 2026-08-29
fix: extract 7z archives that use the ARM64 filter

Marukome0743:
## Check List

  - Link to the issue: #4639

---

Close #4639

## First: you already decided to wait, and I'm not trying to reopen that lightly

In #4871 you said:

> We won't use this version of mholt/archives.
> We'll wait until mholt/archives cuts a new release.
> I don't think we need to rush, since this only affects 7z on windows/arm64.

That was the right call for what was in front of you — that PR used an unreleased pseudo-version of mholt/archives behind a `replace`. **This PR does not.** No `replace`, no pseudo-version, and `mholt/archives` stays on its released `v0.1.5`.

Three things have changed or turned out differently since May. If none of them move you, please just close this — I'll take that as settled and won't raise it again.

## 1. The wait has no end in sight

- `mholt/archives` fixed this on `main` on 2026-05-26 by requiring `sevenzip v1.6.2`.
- Its latest release is still **`v0.1.5` (2025-10-13)**, which requires `sevenzip v1.6.1`.
- The release request in [mholt/archives#71](https://github.com/mholt/archives/issues/71) is from 2026-06-03. Nothing since.

`sevenzip` is an indirect dependency, so Renovate will not bump it on its own. As long as `v0.1.5` is the latest `archives` release, aqua stays on `sevenzip v1.6.1` indefinitely.

## 2. aqua does not fail — it silently produces a broken install

This is the part I think is worth reconsidering. "Only affects 7z on windows/arm64" is accurate, but the failure mode is not an error.

`handleRegularFile` in `pkg/unarchive/archives.go` logs and continues:

```go
reader, err := f.Open()
if err != nil {
	slogerr.WithError(h.logger, err).Warn("open a file")
	return nil
}
```

So every entry using the ARM64 filter is skipped with a warning, `Extract` returns nil, and aqua reports a successful install of a package that is mostly missing.

Measured on `ImageMagick-7.1.2-30-portable-Q16-arm64.7z` by calling `unarchive.Unarchiver.Unarchive` directly:

| `sevenzip` | warnings | extracted |
|---|---|---|
| `v1.6.1` (current `main`) | 8 × `sevenzip: read error: sevenzip: unsupported compression algorithm` | **0 of 8 `.exe`, 1.6 MB** |
| `v1.6.5` (this PR) | 0 | **8 of 8 `.exe`, 234 MB** |

The `-x64` archive of the same release extracts 8 of 8 both before and after, so this is confined to the ARM64-filtered archives.

## 3. The fix is a released dependency, not a workaround

The `archives` fix was only the `sevenzip` version requirement — the `7z.go` diff between `v0.1.5` and `main` is unrelated (match refinement and comments). `sevenzip` registers the ARM64 filter decompressor globally from `init()`, so `archives` picks it up unmodified.

Requiring `sevenzip v1.6.5` directly makes minimal version selection choose it while `archives` stays at `v0.1.5`:

```
$ go list -m github.com/mholt/archives github.com/bodgit/sevenzip
github.com/mholt/archives v0.1.5
github.com/bodgit/sevenzip v1.6.5
```

`go mod tidy` keeps it. The only hand-edited line is:

```diff
-	github.com/bodgit/sevenzip v1.6.1 // indirect
+	github.com/bodgit/sevenzip v1.6.5 // indirect
```

The rest of the `go.mod` diff is what `go mod tidy` pulled in behind it (`brotli`, `klauspost/compress`, `lz4`, `go4.org`, `x/crypto`, `x/text`, and a new `stangelandcl/ppmd`).

When `archives v0.1.6` is eventually released, Renovate will bump `archives` normally and this explicit requirement can be dropped — or just left to age out.

## Verification

```console
$ go vet ./...
$ go test ./... -race -covermode=atomic
$ golangci-lint run
0 issues.
```

Reproduction was done on `darwin/arm64` with aqua built from this branch, extracting the Windows ARM64 archive directly through `pkg/unarchive` (the code path `aqua i` uses), plus a standalone module pinning `archives v0.1.5` with each `sevenzip` version to confirm the bump alone is what changes the outcome.

I'd also be glad to send a follow-up that makes `handleRegularFile` surface these failures instead of warning past them, if you think a truncated-but-successful install is worth turning into an error. Kept out of this PR to keep it to one thing.

---

*AI-assisted: investigated and authored with Claude Code (claude-opus-5). I have reviewed the change and reproduced the before/after measurements myself, and I take responsibility for the content.*

## Summary by CodeRabbit

* **Chores**
  * Updated compression, archive, cryptography, text, and terminal support components.
  * Added support for an additional compression-related component.
  * Updated the Go toolchain for improved compatibility and maintenance.
  * Includes dependency maintenance and compatibility improvements without introducing new user-facing features.


suzuki-shunsuke:
Thank you!

suzuki-shunsuke:
🎉 https://github.com/aquaproj/aqua/releases/tag/v2.63.0

PR #5216, opened 2026-09-09
docs: guide users through persistent PATH setup

### LINKED ISSUE #5205: Clarify how to persist PATH configuration after installing aqua
### Feature Overview

The current installation documentation explains that users need to add aqua's bin directory to `PATH`:

```sh
export PATH="${AQUA_ROOT_DIR:-${XDG_DATA_HOME:-$HOME/.local/share}/aquaproj-aqua}/bin:$PATH"
```

`aqua-installer` also outputs a similar instruction after installation.

However, it may not be obvious to users, especially those who are not familiar with shell configuration, that this setting needs to be persisted in a shell configuration file such as `~/.bashrc`.

If the command is only executed in the current shell, `aqua` will no longer be available from `PATH` after opening a new shell.

I would suggest making the persistent setup more explicit in the installation documentation.

For example, for Bash, the documentation could provide a copy-paste command such as:

```sh
cat >> ~/.bashrc <<'EOF'
export PATH="${AQUA_ROOT_DIR:-${XDG_DATA_HOME:-$HOME/.local/share}/aquaproj-aqua}/bin:$PATH"
EOF

source ~/.bashrc
```

Similarly, the output of `aqua-installer` could explicitly mention that the `export PATH=...` line should be added to the user's shell configuration, rather than only displaying the export command.

For example:

```console
[INFO] Add the following line to your shell configuration (e.g. ~/.bashrc):
export PATH="${AQUA_ROOT_DIR:-${XDG_DATA_HOME:-$HOME/.local/share}/aquaproj-aqua}/bin:$PATH"
```

Currently, the documentation correctly explains that aqua's bin directory must be added to `PATH`, and `aqua-installer` also outputs the required `export PATH=...` command.

However, neither makes it sufficiently explicit that the configuration normally needs to be persisted in a shell configuration file.

I have onboarded several beginners in my community to aqua, and this has consistently been one of the points where they get stuck. In many cases, they successfully install aqua but do not realize that the PATH configuration needs to be persisted in their shell configuration file.

Experienced shell users will likely understand this immediately, but beginners may:

1. run `aqua-installer`,
2. overlook or temporarily execute the displayed `export PATH=...`,
3. open a new shell,
4. find that `aqua` is no longer available.

This can make a successful installation appear broken.

Providing a concrete command for persisting the configuration would make the installation process more beginner-friendly and reduce ambiguity.

### Example Code

For Bash:

```sh
cat >> ~/.bashrc <<'EOF'
export PATH="${AQUA_ROOT_DIR:-${XDG_DATA_HOME:-$HOME/.local/share}/aquaproj-aqua}/bin:$PATH"
EOF

source ~/.bashrc
```

Equivalent examples could be provided for other supported shells where appropriate.

### Note

Related issues:

- aquaproj/aqua-installer#365 addressed outputting the `PATH` setup guide from `aqua-installer`.
- aquaproj/aqua#1670 discussed automatically modifying `.bashrc` / `.zshrc`. The maintainer mentioned preferring to output guidance rather than having a script modify shell configuration automatically.
- aquaproj/aqua#1658 discusses PATH setup guidance for Homebrew.

This request is specifically about making **persistence of the PATH configuration** explicit in the installation documentation and/or installer guidance, rather than merely displaying the `export PATH=...` command.

Shion1305:
Users can finish installing aqua without realizing that PATH must be saved for future terminal sessions. This reorganizes the installation guide into method selection, persistent shell setup, and verification in a new terminal.

- Add responsive installation-method tabs with platform subtitles and a PATH reminder in every local method.
- Suggest the relevant shell tab and provide persistent Bash, Zsh, PowerShell, and Command Prompt instructions; keep Git Bash under Bash and advanced details collapsed.
- Provide separate GitHub Actions, CircleCI Orb, and Dev Container Feature tabs, and link aqua-installer documentation to the setup steps.
- Keep tab selection synchronized with the URL after hydration, reveal hidden panels for anchor navigation, and preserve installation methods in the shell links.

Closes #5205.

## Initial layout preview

output

## Validation

- Website production build passed; it reports existing broken-anchor warnings on four unrelated documentation pages.
- Production Chrome checks passed with no console errors: direct query-string loads, PowerShell/Zsh anchors, hidden-panel TOC links, CI shell links, reloads, Back/Forward (including removal of query parameters), keyboard navigation, and tab scroll preservation. Checked desktop and 412px layouts with no horizontal page overflow.
- Additional production regressions passed: Git Bash heading links expand their collapsed content and scroll after the animation (including reduced motion); clicking an active tab preserves the selected shell, URL and history; invalid method values including `__proto__`, `constructor` and `toString` fall back to Shell Script without React errors.
- Checked generated tab panels and callouts, plus Bash/Zsh snippets in temporary configuration files. Git Bash command syntax was checked with Bash; Windows-specific commands were not executed on Windows.
- `cmdx l`, `cmdx v`, and `cmdx t` passed.
- `git diff --check` passed.

## Check List

  - Link to the issue: https://github.com/aquaproj/aqua/issues/5205

The design and wording were developed through hands-on review; AI assistance was used for implementation and validation.

## Summary by CodeRabbit

- **Documentation**
  - Restructured installation instructions into guided tabs for installation methods and shell-specific PATH setup.
  - Added expanded instructions for Bash, Zsh, PowerShell, and Command Prompt.
  - Added a “Check your setup” step and made binary verification details collapsible.
  - Added dedicated guidance for GitHub Releases, CircleCI Orb, and Dev Container Feature.
  - Updated the aqua installer page with PATH configuration and setup verification guidance.

- **New Features**
  - Added responsive, URL-linked navigation for installation methods and platform setup.
  - Added contextual tips and suggested platform selections to simplify installation.

### DIFF
--- a/website/docs/install.md
+++ b/website/docs/install.md
@@ -1,27 +1,57 @@
 ---
 sidebar_position: 150
+mdx:
+  format: mdx
 ---
+
+import TabItem from '@theme/TabItem';
+import InstallationSetup, {InstallationMethodTabs, InstallationShellTabs, InstallationTabLabel} from '@site/src/components/InstallationSetup';
 
 # Install
 
 aqua is a single binary written in Go.
 
-1. Install the binary `aqua` in `PATH`
-1. Set the environment variable `PATH`
-1. [(Optional) Shell completion](/docs/reference/config/shell-completion)
+For your local terminal, complete all three steps:
+
+1. **Install aqua** using one of the methods below.
+2. **Save the `PATH` setting** in your shell configuration: [Bash](?platform=linux#bash), [Zsh](?platform=macos#zsh), or [Windows](?platform=windows#powershell). Running `export` only in your terminal does not persist the setting.
+3. **Open a new terminal** and run `aqua -v` to check that aqua is still available.
+
+You can also enable [shell completion](/docs/reference/config/shell-completion) after setup.
 
 ## 1. Install the binary `aqua` in `PATH`
 
-- [Homebrew](#homebrew)
-- Windows
-  - [Winget](#winget)
-  - [Scoop](#scoop)
-- [aqua-installer (Shell Script)](/docs/products/aqua-installer#shell-script)
-- [aqua-installer (GitHub Actions)](/docs/products/aqua-installer#github-actions)
-- [CircleCI Orb](/docs/products/circleci-orb-aqua)
-- [go install](#go-install)
-- [Dev Container Feature](https://github.com/aquaproj/devcontainer-features/tree/main/src/aqua-installer)
-- [Download prebuilt binaries from GitHub Releases](#download-prebuilt-binaries-from-github-releases)
+Choose one installation method. For a local terminal, continue to step 2 to save your `PATH` setting.
+
+<InstallationMethodTabs>
+<TabItem value="script" label={<InstallationTabLabel title="Shell Script" detail="Linux / macOS / WSL" />}>
+
+For Linux, macOS, and WSL.
+
+### aqua-installer (Shell Script)
+
+Run the installer, then continue to step 2 below to save your `PATH` setting.
+
+```bash
+curl -sSfL -O https://raw.githubusercontent.com/aquaproj/aqua-installer/v4.0.2/aqua-installer
+echo "98b883756cdd0a6807a8c7623404bfc3bc169275ad9064dc23a6e24ad398f43d  aqua-installer" | sha256sum -c -
+chmod +x aqua-installer
+./aqua-installer
+```
+
+On macOS, use `shasum -a 256 -c -` instead of `sha256sum -c -`.
+See [aqua-installer](/docs/products/aqua-installer#shell-script) for installer options.
+
+:::tip Next: save your PATH setting
+Next, [save the `PATH` setting](#2-set-the-environment-variable-path) so your shell can find tools installed by aqua.
+You still need to complete step 2 to save this PATH setting for new terminal sessions.
+:::
+
+</TabItem>
+
+<TabItem value="homebrew" label={<InstallationTabLabel title="Homebrew" detail="macOS / Linux" />}>
+
+For macOS and Linux with Homebrew installed.
 
 ### Homebrew
 
@@ -39,6 +69,17 @@ Or
 brew install aquaproj/aqua/aqua
 ```
 
+:::tip Next: save your PATH setting
+Next, [save the `PATH` setting](#2-set-the-environment-variable-path) so your shell can find tools installed by aqua.
+You still need to complete step 2 to save this PATH setting for new terminal sessions.
+:::
+
+</TabItem>
+
+<TabItem value="winget" label={<InstallationTabLabel title="Winget" detail="Windows" />}>
+
+For Windows. Run the command in PowerShell or Command Prompt.
+
 ### Winget
 
 From [aqua v2.17.4](https://github.com/aquaproj/aqua/releases/tag/v2.17.4), you can install aqua by [Winget](https://learn.microsoft.com/en-us/windows/package-manager/winget/).
@@ -52,6 +93,17 @@ Due to the mechanism of Winget, it takes a few days at most until we can install
 Everytime a new version is released, we need to send a pull request to [microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs) and wait until the pull request is merged.
 [The list of pull requests](https://github.com/microsoft/winget-pkgs/pulls?q=is%3Aopen+is%3Apr+author%3Asuzuki-shunsuke+aquaproj.aqua+in%3Atitle)
 :::
+
+:::tip Next: save your PATH setting
+Next, [save the `PATH` setting](#2-set-the-environment-variable-path) so your shell can find tools installed by aqua.
+You still need to complete step 2 to save this PATH setting for new terminal sessions.
+:::
+
+</TabItem>
+
+<TabItem value="scoop" label={<InstallationTabLabel title="Scoop" detail="Windows" />}>
+
+For Windows with Scoop installed.
 
 ### Scoop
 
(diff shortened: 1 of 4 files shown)


Shion1305:
Hi @suzuki-shunsuke 👋 
I've reorganized the installation page so that users can easily understand required steps. Hope you like my enhancement 🚀

Shion1305:
@suzuki-shunsuke Some CIs are failing, but this is irrelevant to my changes. Looks like that CI failed to pull a go dependency from Takumi, the GMO Flatt Security's registry, caused the failures.

suzuki-shunsuke:
Thank you for the pull request. The tabbed layout makes the page much more compact, and that part looks good to me. I found some bugs, plus a couple of points I would like to discuss.

Some of the bugs don't reproduce with `npm start`. They need a production build:

```sh
cd website
npm run build && npm run serve   # http://localhost:3000
```

## Bugs

### 1. Hydration mismatch

- `website/src/components/InstallationSetup.js:33`
- `website/src/components/InstallationSetup.js:70`

With `npm run serve` running, open http://localhost:3000/docs/install?method=github-actions in a new tab and look at the DevTools console. A hydration failure is reported:

```
main.40e2281f.js:2 Docusaurus React Root onRecoverableError: Error: Minified React error #418; visit https://reactjs.org/docs/error-decoder.html?invariant=418 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at ua (main.40e2281f.js:2:242662)
    at xl (main.40e2281f.js:2:315405)
    at yc (main.40e2281f.js:2:302963)
    at bc (main.40e2281f.js:2:302935)
    at rc (main.40e2281f.js:2:298016)
    at x (main.40e2281f.js:2:363014)
    at MessagePort.P (main.40e2281f.js:2:363546)
```

React error #418 is "Hydration failed because the initial UI does not match what was rendered on the server". The query string is empty during SSR, so the prerendered HTML always contains steps 2 and 3, while the first client render for `?method=github-actions` returns only a single paragraph.

The shell tabs (`InstallationSetup.js:70`) have the same cause: with a URL such as `?platform=windows#powershell`, `aria-selected` and `hidden` differ from the prerendered HTML.

### 2. Anchor links don't scroll

Open http://localhost:3000/docs/install?platform=windows#powershell directly in a new tab (a full page load, not an in-site navigation). The page stays at the top and `#powershell` has no effect.

In the prerendered HTML the target heading is inside `<div role="tabpanel" hidden>`, and a `display: none` element is not a scroll target. Docusaurus's `scrollAfterNavigation` also does nothing on the initial load because there is no `previousLocation`, so nothing scrolls after hydration switches the tab.

### 3. Dead table of contents entries

Open /docs/install in the production build and click Homebrew / Winget / Scoop / go install / Download prebuilt binaries in the right-hand table of contents. Nothing happens, because the target headings are inside hidden panels.

The same applies to `?method=github-actions` and `?method=others`. There, the `## 3. Check your setup` heading is removed from the DOM, but the table of contents is generated statically from the MDX, so the entry remains.

### 4. The links at the top of the page drop `?method=`, so the URL and the display diverge

The links at `website/docs/install.md:17` are emitted as `href="/docs/install?platform=linux#bash"`, which drops any existing `?method=`.

Docusaurus's Tabs does not re-sync its internal state when the query parameter disappears (`valueToSync` becomes null, so `setSelectedValue` is not called), so the selected tab itself stays as it was. The two actual problems are:

- Reloading, bookmarking or sharing the URL shows a different tab (Shell Script) than the one that was displayed.
- `InstallationSetup` and `InstallationShellTabs` read `method` from the URL rather than from the Tabs state, so they behave as `method=script` as soon as the parameter is gone.

The second one can be seen like this:

1. Open /docs/install
2. Click the GitHub Actions tab (the URL becomes `?method=github-actions`, and step 2 becomes the single "PATH is configured automatically" paragraph)
3. Click the Bash / Zsh / Windows link in the step 2 sentence at the top of the page (that line is outside the tabs, so it is always visible)

The install method tab still shows GitHub Actions, but the full set of local shell tabs is back in step 2, which contradicts the callout in the GitHub Actions tab saying that steps 2 and 3 are not needed.

## Other comments

1. I would rather not have CircleCI Orb and Dev Container Feature grouped together under Others.
2. Prebuilt binaries is labelled Manual download. That is not exactly wrong, but it is not necessarily a manual download. Other platforms such as Jenkins or GitLab may download the prebuilt binaries too.

Shion1305:
@suzuki-shunsuke  Thanks for the review! I've addressed your feedback and verified the changes with npm run build and npm run serve. Could you take another look when you have a chance 🙏

Shion1305:
@suzuki-shunsuke Lmk if there’s anything I can improve in this PR. Happy to make further changes!

suzuki-shunsuke:
Thank you for the quick turnaround. I checked the latest revision and the four bugs I reported are fixed.

I would like to step back and reconsider the design, though. The tab design came from the issue and you implemented it faithfully, so this is a change of direction on my side as the maintainer, not a problem with your work. I am sorry for raising it this late.

## Maintaining the JavaScript

`InstallationSetup.js` is now about 160 lines, and it depends on Docusaurus internals: the `data-collapsed` attribute of `Details`, the height transition of `Collapsible`, and the fact that `scrollAfterNavigation` does nothing on the initial load. If a Docusaurus update changes any of those, the page breaks quietly. The build still passes and the page still renders, and we only find out when someone reports that an anchor no longer works. I do not want to own that kind of code in the documentation site.

## What the issue actually needs

The problem is that people install aqua and never persist `PATH`. Looking at the current page again:

- `Set the environment variable PATH` is already in the list at the top of the page and is already a section heading. It is not hidden.
- What is missing is that the section never says the line belongs in `~/.bashrc`, `~/.zshrc` and so on. Saying that is genuinely helpful, and it is the part of this PR I would like to keep.
- I would rather not add `>> ~/.bashrc` style commands. Running one twice appends the line twice, and it adds a redundant entry for anyone who has already configured `PATH`. Showing the line and naming the file it goes in is enough.

## Tabs

- The benefit of the tabs is a tidier page.
- The page is not that big to begin with: `website/docs/install.md` is 260 lines on `main`, and 99 of them are the `Verify downloaded binaries from GitHub Releases` section.
- The complexity the tabs bring is not proportionate to that benefit.
- If we want a tidier page, `<details>` is enough. Rather than one `<details>` around the whole `Verify downloaded binaries from GitHub Releases` section, I would put one around each tool in it (Cosign, slsa-verifier, GitHub CLI), so that the three options stay visible as a list and only the commands and their output collapse. The PowerShell and Command Prompt commands could be wrapped the same way.

## What I would like to ask for

Please keep the documentation changes:

- naming the configuration file the `export` line belongs in
- the "Next: save your PATH setting" callout after each installation method
- the new step 3 that has the reader open a new terminal and run `aqua -v`

and drop `InstallationSetup.js`, `InstallationSetup.module.css` and the tabs, using `<details>` around the long command blocks instead.

I know this discards a lot of the work in this PR, and I appreciate the time you put into it. Let me know if you would prefer that I take it from here.

